PR #1, opened 2022-10-24
0001 add login screen

realityexpander:
Add basic login & registration functionality, validation & screens.
Setup Hilt and navigation.
Does not use resources for strings.
Does not hit real API or Database. 
Uses fakes, but implements all necessary functionality. 
No styling, except basic compose UI.
Prepped structure for implementing Database and API next.


app/src/main/java/com/realityexpander/tasky/presentation/login_screen/LoginScreen.kt
```diff
@@ -1,22 +1,174 @@
...
-    Text(text = "Login")
+    val loginState by viewModel.loginState.collectAsState()
+    val scope = rememberCoroutineScope()
+    val focusManager = LocalFocusManager.current
+
+    fun performLogin() {
+        scope.launch {
+            viewModel.onEvent(LoginEvent.ValidateEmail(loginState.email))
+            viewModel.onEvent(LoginEvent.ValidatePassword(loginState.password))
+            viewModel.onEvent(LoginEvent.Login(loginState.email, loginState.password))
```
philipplackner, line 44: *Personal opinion alert*
To me, it should rather be the job of the ViewModel to decide what needs to be done in which order. When the user clicks on Login, all the UI knows is that exact action. The ViewModel could then validate the fields before attempting to make the network call. 
But, if you disagree with this, feel free to leave this, just sharing my view here.

app/src/main/java/com/realityexpander/tasky/presentation/login_screen/LoginScreen.kt
```diff
@@ -1,22 +1,174 @@
...
 ) {
+    BackHandler(true) { /* We want to disable back clicks */ }
 
-    Text(text = "Login")
+    val loginState by viewModel.loginState.collectAsState()
+    val scope = rememberCoroutineScope()
+    val focusManager = LocalFocusManager.current
+
+    fun performLogin() {
+        scope.launch {
```
philipplackner, line 41: Is there a reason you chose the UI scope over `viewModelScope`? 
What if there is a config change during execution of this function?

app/src/main/java/com/realityexpander/tasky/presentation/login_screen/LoginScreen.kt
```diff
@@ -1,22 +1,174 @@
...
+        }
+        Spacer(modifier = Modifier.height(32.dp))
+
+        // REGISTER BUTTON
+        Text(text = "Not a member? Sign up",
+            color = Color.Cyan,
+            modifier = Modifier
+                .align(alignment = Alignment.CenterHorizontally)
+                .clickable(onClick = {
+                    navigator.clearBackStack("LoginScreenDestination")
```
philipplackner, line 136: I think with this library, you can also pass `LoginScreenDestination` as an object instead of as a string (not 100% sure here). That would have the benefit that if the name of `LoginScreen` would change, the compiler would notify you here, while with a string it would silently break.

app/src/main/java/com/realityexpander/tasky/presentation/login_screen/LoginScreen.kt
```diff
@@ -1,22 +1,174 @@
...
+                text = "Error: ${loginState.errorMessage}",
+                color = Color.Red,
+            )
+            Spacer(modifier = Modifier.height(8.dp))
+        }
+        if(loginState.isLoggedIn) {
+            Text(text = "Logged in!")
+            Spacer(modifier = Modifier.height(8.dp))
+        }
+        if(loginState.statusMessage.isNotEmpty()) {
```
philipplackner, line 168: I personally prefer `null` values if something is not existent (that's exactly what nulls are for). Having a non-null string that is empty doesn't make it fully clear if there is a status message or not.

The only scenario where I use empty strings as default values would be text field values, since in that case it does make sense to me to have an empty string.

app/src/main/java/com/realityexpander/tasky/presentation/login_screen/LoginViewModel.kt
```diff
@@ -2,37 +2,144 @@ package com.realityexpander.tasky.presentation.login_screen
...
+    private val validatePassword: IValidatePassword,
     private val savedStateHandle: SavedStateHandle,
 ) : ViewModel() {
 
-    suspend fun login(email: String, password: String) {
+    private val _loginState = MutableStateFlow<LoginState>(LoginState())
+    val loginState: StateFlow<LoginState> = _loginState.asStateFlow()
+
+    private val email: String = savedStateHandle[SAVED_STATE_EMAIL] ?: ""
+    private val password: String = savedStateHandle[SAVED_STATE_PASSWORD] ?: ""
```
philipplackner, line 31: Why are these the only fields you want to restore after process death? What about validation errors or status messages for example?

app/src/main/java/com/realityexpander/tasky/presentation/login_screen/LoginViewModel.kt
```diff
@@ -2,37 +2,144 @@ package com.realityexpander.tasky.presentation.login_screen
...
-    ),
+    private val authRepository: IAuthRepository,
+    private val validateEmail: IValidateEmail,
+    private val validatePassword: IValidatePassword,
     private val savedStateHandle: SavedStateHandle,
 ) : ViewModel() {
 
-    suspend fun login(email: String, password: String) {
+    private val _loginState = MutableStateFlow<LoginState>(LoginState())
+    val loginState: StateFlow<LoginState> = _loginState.asStateFlow()
```
philipplackner, line 28: Something you could do to avoid needing to update `email` and `password` in `savedStateHandle` every time would be this:
```
val loginState = _loginState.onEach { state ->
    savedStateHandle[SAVED_STATE_EMAIL] = state.email
    savedStateHandle[SAVED_STATE_PASSWORD] = state.password
}.stateIn(viewModelScope, SharingStarted.WhileSubscribed(5000), LoginState())
```

Then, no matter where these fields could change, you can be sure that savedStateHandle is up-to-date.

You could also adjust this a bit to automatically validate email, password and corresponding error messages whenever they change (using the `map` operator)

app/src/main/java/com/realityexpander/tasky/presentation/register_screen/RegisterScreen.kt
```diff
@@ -1,14 +1,211 @@
...
 fun RegisterScreen(
-    vieModel: RegisterViewModel = hiltViewModel()
+    email: String? = null,
+    password: String? = null,
+    confirmPassword: String? = null,
+    navigator: DestinationsNavigator,
+    viewModel: RegisterViewModel = hiltViewModel(),
 ) {
-    Text(text = "Register")
+    BackHandler(true) { /* We want to disable back clicks */ }
```
philipplackner, line 34: Any reason for that?

app/src/test/java/com/realityexpander/tasky/domain/validation/IValidatePasswordTest.kt
```diff
@@ -0,0 +1,45 @@
...
+import com.realityexpander.tasky.data.validation.ValidateEmailImpl
+import com.realityexpander.tasky.data.validation.ValidatePasswordImpl
+import junit.framework.TestCase.assertFalse
+import junit.framework.TestCase.assertTrue
+import org.junit.Test
+
+
+class IValidatePasswordTest {
+
+    private val validatePassword = ValidatePasswordImpl()
```
philipplackner, line 12: Why is the test class name based on the interface, but you test the implementation? Shouldn't it be `ValidatePasswordImplTest` instead`

philipplackner (approved):

app/src/main/java/com/realityexpander/tasky/data/repository/AuthRepositoryImpl.kt
```diff
@@ -14,7 +14,7 @@ class AuthRepositoryImpl(
 ): IAuthRepository {
```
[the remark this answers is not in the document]
realityexpander: It was used quite a bit back in earlier android days, it just makes it super obvious when using an interface versus an Impl. I like it because I dont have to check what it is, i can just look at the name and know its a concrete object or an interface.

app/src/main/java/com/realityexpander/tasky/domain/validation/IValidatePassword.kt
```diff
@@ -0,0 +1,5 @@
+package com.realityexpander.tasky.domain.validation
+
+interface IValidatePassword {
+    fun validatePassword(password: String): Boolean
```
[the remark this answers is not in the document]
realityexpander: Good point. I will remove this abstraction.

app/src/main/java/com/realityexpander/tasky/presentation/components/EmailField.kt
```diff
@@ -0,0 +1,73 @@
...
+import androidx.compose.ui.text.input.PasswordVisualTransformation
+import androidx.compose.ui.text.input.VisualTransformation
+import com.realityexpander.tasky.presentation.register_screen.RegisterEvent
+import kotlinx.coroutines.launch
+
+@Composable
+fun EmailField(
+    email: String,
+    label: String = "Email",
+    placeholder: String = "Enter your Email",
```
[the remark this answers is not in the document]
realityexpander: OK, i will implement your string resource class you showed in a previous video on your channel.

app/src/main/java/com/realityexpander/tasky/presentation/components/EmailField.kt
```diff
@@ -0,0 +1,73 @@
...
+    isError: Boolean,
+    onEmailChange: (String) -> Unit,
+    keyboardOptions: KeyboardOptions =
+        KeyboardOptions(keyboardType = KeyboardType.Email, imeAction = ImeAction.Next),
+    keyboardActions: KeyboardActions? = null,
```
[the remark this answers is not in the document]
realityexpander: I wanted to make this as domain specific to just using the composable for an email field.

But I see your point, and will make it more generic. :)

app/src/main/java/com/realityexpander/tasky/presentation/components/PasswordField.kt
```diff
@@ -0,0 +1,77 @@
...
+    placeholder: String = "Enter your Password",
+    isError: Boolean,
+    isPasswordVisible: Boolean = false,
+    clickTogglePasswordVisibility: () -> Unit,
+    onPasswordChange: (String) -> Unit,
+    imeAction: ImeAction = ImeAction.Done,
+    keyboardOptions: KeyboardOptions =
+        KeyboardOptions(keyboardType = KeyboardType.Password, imeAction = imeAction),
+    keyboardActions: KeyboardActions? = null,
+    doneAction : () -> Unit = {},
```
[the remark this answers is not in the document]
realityexpander: OK, I will add the modifier. 👍🏼

app/src/main/java/com/realityexpander/tasky/presentation/components/PasswordField.kt
```diff
@@ -0,0 +1,77 @@
...
+    )
+
+    OutlinedTextField(
+        value = password,
+        singleLine = true,
+        onValueChange = onPasswordChange,
+        isError = isError,
+        label = { Text(text = label) },
+        placeholder = { Text(text = placeholder) },
+        modifier = Modifier.fillMaxWidth(1f),
```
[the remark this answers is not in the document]
realityexpander: OK, I will add passing in the modifier. Good catch. 👍🏼

app/src/main/java/com/realityexpander/tasky/presentation/login_screen/LoginScreen.kt
realityexpander: I can see the value in this.

OK, i will move this to the viewmodel and do the validations before the `LoginEvent.login`

app/src/main/java/com/realityexpander/tasky/presentation/login_screen/LoginScreen.kt
realityexpander: Good catch, I was wondering about this. I will move any viewModel specific coroutines to the ViewModel's `.sendEvent` function. I will keep Compose-specific scopes in the Composable.

💯

app/src/main/java/com/realityexpander/tasky/presentation/login_screen/LoginScreen.kt
realityexpander: OK, i will change to the other way of clearing the backstack, without using hard-coded strings.

😺

app/src/main/java/com/realityexpander/tasky/presentation/login_screen/LoginScreen.kt
realityexpander: OK this makes sense. I will convert these "no message" strings to nullables, and check for `null` instead of `isNotEmpty` 

😀

app/src/main/java/com/realityexpander/tasky/presentation/login_screen/LoginViewModel.kt
realityexpander: Good point, i will include those fields for process death. 🧟‍♂️

app/src/main/java/com/realityexpander/tasky/presentation/login_screen/LoginViewModel.kt
realityexpander: OK, I wondered about this. I will make this change, and thanks for using my example in the call! 😄

app/src/main/java/com/realityexpander/tasky/presentation/register_screen/RegisterScreen.kt
realityexpander: When I implement the splash screen, i don't want the user to go back to the splash screen. I haven't tried the splash screen processing yet, so I assumed that allowing the back button would return to the splash screen... maybe it doesn't?

And I have a button on the screen to flip back and forth between `Sign-In` and `Register`, so I also thought it kept the from returning to the previous screen.

2 questions:

1. So, should I allow the user to use the back button? It wont return to the splash screen, correct?
3. And when the user successfully logs in, should I clear the backstack for the `LoginScreen` and `RegistrationScreen`?

app/src/test/java/com/realityexpander/tasky/domain/validation/IValidatePasswordTest.kt
realityexpander: Good call, I wasnt sure if there was a android-specific way to do password validation, so I was leaving this open for further research.

But it looks like there isnt (im just using Regex), so I will make this a concrete implementation and remove the interface abstraction.

app/src/main/java/com/realityexpander/tasky/data/validation/ValidateEmailImpl.kt
```diff
@@ -7,6 +7,8 @@ class ValidateEmailImpl(
     private val emailMatcher: IEmailMatcher = EmailMatcherImpl()
 ): IValidateEmail {
     override fun validateEmail(email: String): Boolean {
+        if(email.isEmpty()) return false
+
         return emailMatcher.matches(email)
```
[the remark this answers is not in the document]
realityexpander: OK, this is a good suggestion! 🤠

app/src/main/java/com/realityexpander/tasky/presentation/register_screen/RegisterScreen.kt
realityexpander: I found a way to force only navigation between the login and registraion. If the user wants to quit, he will have to force quit the app currently

PR #5, opened 2022-10-29
0003 login use actual tasky api

realityexpander:
Sorry I made this PR larger than I would have liked, but I was on a roll! I will attempt to make the next one smaller! 

PLEASE DON'T BEAT ME!

## Changes / Additions
- Now accesses the Prod Tasky API endpoint via Retrofit
- Add Logging interceptor
- Add `x-api-key` header interceptor
- improved fake API and Dao implementations
- Added `fake` and `prod` implementations in Hilt
- Setup ability to switch between `Prod` and `Fake` implementations easily
- Simplified UI previews to not use ViewModel instantiation, previews are much faster and easier to build/test
- `AuthRepository` now shares validation code (instead of adding it twice like i was doing before)
- Use `api.properties` file to hide API key (not checked into repo)
- Added `release` variant and automatically turns off `OkHttp3` logging for `release` builds
- Added regex implementation of `EmailMatcher` to match validation on the server
- Rearranged folders for clarity
- Improved API error handling
- Added `ExitActivity` to cleanly quit the app.
- Update and fix tests.

## Question 1
Where to put RetroFit request/response data classes (what packages)?

## Question 2
Is using an `ExitActivity` the preferred way to quit the app? I found a lot of conflicting opinions on this.

Here:
https://github.com/realityexpander/Tasky/blob/3cb6db4e96631e0abc0ca9c3643ba3f84dfde88c/app/src/main/java/com/realityexpander/tasky/ExitActivity.kt

## Question 3
What is the optimal package setup? Should I have a package for `authentication flow` and another for `management/main-app`? I dont want to split the app into modules yet, but i would like to set the project up to do that at some point.

Please review my package structure.

## Question 4
Should I hard-code API response codes, as I am doing in `AuthApiImpl.kt`? Or should i make them more general (all 4XX's do one error, all 5XX's do another error)?

Here:
https://github.com/realityexpander/Tasky/blob/3cb6db4e96631e0abc0ca9c3643ba3f84dfde88c/app/src/main/java/com/realityexpander/tasky/data/repository/remote/authApiImpls/AuthApiImpl.kt#L40

## Question 4
In `AppModule.kt`: (check comments in that file)
- How to add `TaskApi` to the `authApi` interface without using `@Provides`? 
- Is there a way to have Hilt instantiate `TaskyApi` using `@Binds`?
  
Or is this the only way to instantiate a dependency is to use `@Provides` when creating an
     implementation of an interface that requires a dependency?

Here:
https://github.com/realityexpander/Tasky/blob/d0d9a026cf508de6a66823f1b6913458f5e9fd5b/app/src/main/java/com/realityexpander/tasky/di/AppModule.kt#L138

```
    @Provides
    @Singleton
    @AuthRepositoryProd_AuthApiProd_AuthDaoFake
    fun provideAuthRepositoryProd(
//        @AuthApiProdUsingBinds authApi: IAuthApi,  // How to add `TaskApi` to the `authApi`? (instead of using @AuthApiProdUsingProvides above)
        @AuthApiProdUsingProvides authApi: IAuthApi,
        @AuthDaoFakeUsingBinds authDao: IAuthDao,
        validateUsername: ValidateUsername,
        @Named("ValidateEmailRegex") validateEmail: IValidateEmail,
        validatePassword: ValidatePassword
    ): IAuthRepository =
        AuthRepositoryImpl(
            authApi = authApi,
            authDao = authDao, // use fake repo until we implement database
            validateUsername = validateUsername,
            validateEmail = validateEmail,
            validatePassword = validatePassword
        )
```


app/src/main/java/com/realityexpander/tasky/di/AppModule.kt
```diff
@@ -1,38 +1,103 @@
...
+    fun provideTaskyApi(): TaskyApi {
+
+        val xApiKeyHeader = Interceptor { chain ->
+            val request = chain.request().newBuilder()
+                .addHeader("x-api-key", API_KEY)
+                .build()
+            chain.proceed(request)
+        }
+
+        val client = if(BuildConfig.BUILD_TYPE == "debug") {
```
philipplackner, line 50: There's also a field `BuildConfig.DEBUG` which you can directly use for such checks

app/src/main/java/com/realityexpander/tasky/domain/validation/validateEmail/EmailMatcherRegexImpl.kt
```diff
@@ -0,0 +1,10 @@
+package com.realityexpander.tasky.domain.validation.validateEmail
+
+// Matches how the email is validated on the server
+class EmailMatcherRegexImpl: IEmailMatcher {
+    override fun matches(email: String): Boolean {
+        // use regex to validate email
+        val regex = Regex("^[A-Z0-9._%+-]+@[A-Z0-9.-]+\\.[A-Z]{2,6}$", RegexOption.IGNORE_CASE)
+        return regex.matches(email)
+    }
+}
```
philipplackner, line 10: In this case you could also put this matching logic in the email use case directly, since there's nothing Android specific about that which you need to abstract here.

app/src/test/java/com/realityexpander/tasky/domain/validation/IValidateEmailTest.kt
```diff
@@ -11,30 +12,56 @@ class IValidateEmailTest {
...
         validateEmail = ValidateEmailImpl(emailMatcherFakeImpl)
 
-        // ACT
-        val result = validateEmail.validate("chris@demo.com")
+        // ACT / ASSERT
+        assertTrue(validateEmail.validate("chris@demo.com")) // valid email
 
-        // ASSERT
-        assertTrue(result)
     }
```
philipplackner, line 26: What's the point of adding that test case if you actually just test your fake implementation?

app/src/main/java/com/realityexpander/tasky/di/AppModule.kt
```diff
@@ -42,4 +107,100 @@ object AppModule {
...
+                authApi = provideAuthApiProd(
+                    provideTaskyApi()
+                ),
+                authDao = AuthDaoFakeImpl(),  // why cant we use the implementation from @Binds here?
+                validateUsername = provideValidateUsername(),
+                validatePassword = provideValidatePassword(),
+                validateEmail = provideValidateEmailRegex(),
+            )
+        }
+    }
```
philipplackner, line 132: Idk, if I add `@AuthDaoFakeUsingBinds fakeDao: IAuthDao` as a parameter here, your app launches. Looks fine to me.

philipplackner:
Answers to your questions:
1. I like to package by context, so for example you'd have a package agenda->data->event in which you can put your event related network models. agenda->data->reminder for reminder related etc. This comes with more packages, but is more organized than putting it all in one big models/requests/responses package. I also noticed, you use layer-based packaging on the root of your project which I'd personally avoid. If your whole app is put into just 3 packages, these packages will get ultra large over time if you have a big app. Instead, I'd recommend to check the video about the right package structure in the video portal where I talk about having feature root packages where each is divided into presentation, domain and data.
2. Interesting, I've never seen this before, but also don't really see the point 🤔 I personally usually use a single activity architecture, so you could just call `finish()` from anywhere in your code to exit the app. And if you want to finish the whole app, you'd need to call `exitProcess(0)` afterwards as well AFAIK.
3. Answered in (1) already, but in your current project structure of this branch I only see presentation, domain, data and common root packages 🤔  Anyways, for this app I'd do it exactly as you mentioned in your question: 2 feature packages `auth` and `agenda` or similar and a `core/common` package for the shared stuff.
4. Usually, I use a few generic API response codes. It's just important that the user knows if it's their fault or the fault of the app/server. And I'd make an exception for 401 if the user is not authenticated.
5. Not sure why you use 2 interfaces for the same API? Couldn't you make it all dependent on the Retrofit interface?

realityexpander:
1. OK, i will rearrange packages by feature with a core package for all things common. I watched the video, and now it makes sense.
2. OK, i used the `finish()/exitProcess(0)` in one attempt, but then saw comments is SO about that may cause problems. They didnt say what exactly... so I went with the "ExitActivity" approach as a safe alternative. I will implement your suggestion.
3. I will restructure the app packages as per your suggestion.
4. OK, i will make the API response codes more generic moving forward.
5. OK, i will use one retrofit instance for the API.

app/src/main/java/com/realityexpander/tasky/di/AppModule.kt
realityexpander: My question is specifically: For `@Bind`ing an interface, is there a way to pass a parameter using the `@Binds` annotation, or is the only way to pass an argument to an interface is to have hilt instantiate it with a `@Provides` and pass the parameter there?

app/src/test/java/com/realityexpander/tasky/domain/validation/IValidateEmailTest.kt
realityexpander: I just want to be sure my fake implementation is actually working properly. Is this a bad idea?

app/src/main/java/com/realityexpander/tasky/domain/validation/validateEmail/EmailMatcherRegexImpl.kt
realityexpander: OK, because the Android-specific implementation does not match the implementation on the server. 

Which is why I had to include this Regex version.

I will remove the Android-specific implementation.

app/src/main/java/com/realityexpander/tasky/di/AppModule.kt
realityexpander: I will use that field instead of checking the build string.

app/src/main/java/com/realityexpander/tasky/data/repository/remote/authApiImpls/AuthApiFakeImpl.kt
```diff
@@ -0,0 +1,104 @@
...
+            )
+    }
+
+    override suspend fun login(
+        email: Email,
+        password: Password
+    ): AuthInfoDTO {
+        if (email.isBlank() || password.isBlank()) {
+            throw Exceptions.LoginException("Invalid email or password")
+        }
```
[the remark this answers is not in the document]
realityexpander: I was thinking this is a basic sanity check. I will remove this check.

app/src/main/java/com/realityexpander/tasky/data/repository/authRepositoryImpls/AuthRepositoryFakeImpl.kt
```diff
@@ -12,17 +17,20 @@ import com.realityexpander.tasky.domain.validation.ValidateUsername
...
 ): IAuthRepository {
 
-    override suspend fun login(email: String, password: String): AuthToken {
+    override suspend fun login(email: String, password: String): AuthInfo {
         if(!validateEmail.validate(email)) {
             throw Exceptions.InvalidEmailException()
         }
+        if(!validatePassword.validate(password)) {
+            throw Exceptions.InvalidPasswordException()
+        }
```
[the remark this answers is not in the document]
realityexpander: OK I will implement this suggestion. I can see the value in this approach.

app/src/main/java/com/realityexpander/tasky/data/repository/authRepositoryImpls/AuthRepositoryImpl.kt
```diff
@@ -0,0 +1,85 @@
...
+            authApi.login(email, password)
+        } catch (e: Exceptions.LoginException) {
+            throw e
+        } catch (e: Exceptions.LoginNetworkException) {
+            throw e
+        } catch (e: Exceptions.WrongPasswordException) {
+            throw e
+        } catch (e: Exception) {
+            throw Exceptions.UnknownErrorException(e.message)
+        }
```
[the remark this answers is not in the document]
realityexpander: OK, i will implement this change.

app/src/main/java/com/realityexpander/tasky/data/repository/authRepositoryImpls/AuthRepositoryImpl.kt
```diff
@@ -0,0 +1,85 @@
...
+            throw e
+        } catch (e: Exceptions.RegisterNetworkException) {
+            throw e
+        } catch (e: Exceptions.EmailAlreadyExistsException) {
+            throw e
+        } catch (e: Exceptions.UnknownErrorException) {
+            throw e
+        } catch (e: Exception) {
+            throw Exceptions.UnknownErrorException(e.message)
+        }
```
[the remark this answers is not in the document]
realityexpander: OK I will implement this change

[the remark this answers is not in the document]
realityexpander: Do we also add this ` catch(e: CancellationException) { throw e }` to the `AuthApiImpl`? When we surround the Retrofit call to the Api with a `try-catch` block?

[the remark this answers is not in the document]
philipplackner: Both works. And you only need to do this if you catch general exceptions in a try block that executes a suspending function. If you only catch specific exceptions (and not cancellation ones therefore), you don't need this

app/src/main/java/com/realityexpander/tasky/di/AppModule.kt
philipplackner: If your `AuthApiImpl` has an inject constructor that takes a `TaskyApi`, Hilt should actually automatically create that with your provided API interface. If there are multiple, it will complain (not sure if you can use `@Named` here in Binds). 
Either way, you also don't do anything wrong with using Provides for everything. Generates a bit more code, but has no performance drawbacks.

PR #6, opened 2022-10-29
0004 add splash screen

realityexpander:
- Added splash screen
- Modified android splash screen to not show logo when coming back from stopped/killed state
- Added stub Agenda screen
- Using SSOT for colors for XML-based splash and secondary color in MaterialTheme using Resource Color
- Added runConfig to rebuild the NavGraph generated code
- Added TextEditField TextStyle parameter (it was not carrying the theme over for some reason)

### DIFF
--- a/app/src/main/java/com/realityexpander/tasky/MainActivity.kt
+++ b/app/src/main/java/com/realityexpander/tasky/MainActivity.kt
@@ -64,10 +64,10 @@ fun SplashScreen(
         } else {
             navigator.navigate(
                 LoginScreenDestination(
-                    username = "Chris Athanas",
-                    email = "chris@demo.com",
-                    password = "Password1",
-                    confirmPassword = "Password1",
+                    username = "Chris Athanas",     // TESTING ONLY
+                    email = "chris@demo.com",       // TESTING ONLY
+                    password = "Password1",         // TESTING ONLY
+                    confirmPassword = "Password1",  // TESTING ONLY
                 )
             ) {
                 launchSingleTop = true
--- a/app/src/main/java/com/realityexpander/tasky/MainActivity.kt
+++ b/app/src/main/java/com/realityexpander/tasky/MainActivity.kt
@@ -1,24 +1,18 @@
 package com.realityexpander.tasky
 
 import android.os.Bundle
-import android.os.Debug.waitForDebugger
-import android.window.SplashScreen
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.compose.foundation.Image
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.*
-import androidx.compose.material.MaterialTheme
 import androidx.compose.material.Surface
-import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
-import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.res.colorResource
 import androidx.compose.ui.res.painterResource
-import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.IntOffset
 import androidx.compose.ui.unit.dp
@@ -27,11 +21,9 @@ import com.ramcosta.composedestinations.annotation.Destination
 import com.ramcosta.composedestinations.annotation.RootNavGraph
 import com.ramcosta.composedestinations.navigation.DestinationsNavigator
 import com.realityexpander.tasky.destinations.LoginScreenDestination
-import com.realityexpander.tasky.destinations.RegisterScreenDestination
 import com.realityexpander.tasky.presentation.ui.theme.TaskyTheme
 import dagger.hilt.android.AndroidEntryPoint
 import kotlinx.coroutines.delay
-import kotlinx.coroutines.launch
 
 @AndroidEntryPoint
 class MainActivity : ComponentActivity() {
@@ -45,7 +37,7 @@ class MainActivity : ComponentActivity() {
             TaskyTheme {
                 Surface(
                     modifier = Modifier.fillMaxSize(),
-                    color = colorResource(id = R.color.green_tasky)
+                    color = colorResource(id = R.color.tasky_green)
                 ) {
                     DestinationsNavHost(navGraph = NavGraphs.root)
                 }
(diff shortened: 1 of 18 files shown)


app/src/main/java/com/realityexpander/tasky/MainActivity.kt
```diff
@@ -40,6 +48,53 @@ class MainActivity : ComponentActivity() {
     }
 }
 
+@Composable
+@Destination
+@RootNavGraph(start = true)
+fun SplashScreen(
```
philipplackner, line 54: I'd avoid using your own Splash screens if it's not a very complex one. Android already offers a built in splash screen API which will always show a splash screen for devices running on 11+. If you implement your own one, it might show 2 splash screens on some devices.
Android's API also allows you to keep the splash screen visible until a certain condition is met (like the user being authenticated). In this video I explain it:
https://www.youtube.com/watch?v=Loo4i5IrZ4Y

app/src/main/java/com/realityexpander/tasky/presentation/agenda_screen/AgendaEvent.kt
```diff
@@ -0,0 +1,44 @@
...
+    object ValidateEmail : AgendaEvent()
+    object ValidatePassword : AgendaEvent()
+    object ValidateConfirmPassword : AgendaEvent()
+    object ValidatePasswordsMatch : AgendaEvent()
+
+    data class IsValidUsername(val isValid: Boolean) : AgendaEvent()
+    data class IsValidEmail(val isValid: Boolean) : AgendaEvent()
+    data class IsValidPassword(val isValid: Boolean) : AgendaEvent()
+    data class IsValidConfirmPassword(val isValid: Boolean) : AgendaEvent()
+    data class IsPasswordsMatch(val isMatch: Boolean) : AgendaEvent()
```
philipplackner, line 24: These classes all sound like something you'd get a result from(the resulting boolean). But in the end, such events just do something or trigger some kind of action, so naming-wise I'd prefer something like `ValidateUsername` or `CheckIfPasswordsMatch` here to make it clear that the UI can trigger such functionality.

app/src/main/java/com/realityexpander/tasky/presentation/agenda_screen/AgendaEvent.kt
```diff
@@ -0,0 +1,44 @@
+package com.realityexpander.tasky.presentation.agenda_screen
+
+import com.realityexpander.tasky.presentation.util.UiText
+
+sealed class AgendaEvent {
```
philipplackner, line 5: Could also be a `sealed interface` if it doesn't need any constructor arguments. That way it'll occupy less memory.

app/src/main/java/com/realityexpander/tasky/presentation/agenda_screen/AgendaState.kt
```diff
@@ -0,0 +1,32 @@
...
+    val password: String = "",
+    val confirmPassword: String = "",
+    val isPasswordVisible: Boolean = false,
+
+    val isLoading: Boolean = false,
+    val isLoggedIn: Boolean = false,
+    val isError: Boolean = false,
+
+    val isInvalidUsername: Boolean = false,
+    val isShowInvalidUsernameMessage: Boolean = false,
```
philipplackner, line 17: Could it make sense to combine these into one field?
```
val invalidUsernameMessage: String? = null

// --------------------------------------

if(invalidUsernameMessage == null) {
    // No error and don't show message
} else {
    // Show message and there is an error
}
```

app/src/main/java/com/realityexpander/tasky/presentation/agenda_screen/AgendaState.kt
```diff
@@ -0,0 +1,32 @@
...
+    val isInvalidPassword: Boolean = false,
+    val isShowInvalidPasswordMessage: Boolean = false,
+
+    val isInvalidConfirmPassword: Boolean = false,
+    val isShowInvalidConfirmPasswordMessage: Boolean = false,
+
+    val isPasswordsMatch: Boolean = true,
+
+    val errorMessage: UiText = UiText.None,
+    val statusMessage: UiText = UiText.None,
```
philipplackner, line 31: Also, here I'd rather work with nulls to indicate there is no error message. An empty error message still means an existing error to me.

philipplackner:
Yes, if you really feel the need to need that many headlines, that's a good way to solve it IMO :)

app/src/main/java/com/realityexpander/tasky/presentation/agenda_screen/AgendaEvent.kt
realityexpander: OK, I will implement this suggestion. I see I was over complicating things here.

app/src/main/java/com/realityexpander/tasky/presentation/agenda_screen/AgendaEvent.kt
realityexpander: Pretty neat. This is my first time using a sealed interface. Very interesting.

OK, i will implement this!

app/src/main/java/com/realityexpander/tasky/presentation/agenda_screen/AgendaState.kt
realityexpander: I have them in two separate fields because on is for while the user is typing, and the other is after they attempt to login/register and I show them the "red error" message. I only want to show the "red error" once they have ignored the "invalid entry" icon in the field. 

So basically I am tracking 2 states: 
1. While the user types, a small icon will show the "check" or the "warning" icon.
2. When the user attempts to login/register, they will see the error messages below the fields, letting them know what is wrong and how to fix it.

app/src/main/java/com/realityexpander/tasky/presentation/agenda_screen/AgendaState.kt
realityexpander: Good idea. I will convert these to nulls.

app/src/main/java/com/realityexpander/tasky/MainActivity.kt
realityexpander: OK, i see this method is much easier than what I was doing. But Im still curious how the two screen would show on some devices, as im just using the `DestinationsNavhost` and navigating from screen to screen.

app/src/main/java/com/realityexpander/tasky/presentation/agenda_screen/AgendaState.kt
realityexpander: Is there a better way to handle this?

PR #8, opened 2022-10-31
0006 update for comments pr0003 0004

realityexpander:
I am trying to make these smaller, but also make them confined to things that make sense. 

- Implemented suggestions from PR-0003 to PR-0005
- Added Android 12 version of splash screen
- simplified ViewModel events
- Added Api call to MainActivity to validate token upon launch
- Moved storage of AuthToken to authDao
- Simplified splashScreen
- Improved and simplified AppModule for hilt
- Getting Authorization AuthToken from the `AuthDao` and adding it to the header for all API calls in the `interceptor`


app/src/main/java/com/realityexpander/tasky/core/common/settings/AppSettings.kt
```diff
@@ -0,0 +1,24 @@
+package com.realityexpander.tasky.core.common.settings
```
philipplackner, line 1: `core.common` seems duplicated since `core` and `common` usually refer to the same thing. Settings are data-related and should therefore go in `core->data`

philipplackner (approved):

app/src/main/java/com/realityexpander/tasky/core/common/settings/AppSettings.kt
realityexpander: OK I will remove the extraneous `common` package, and move the settings.

PR #9, opened 2022-11-02
0007 add logout button

realityexpander:
I am working hard to make these smaller!

- Add logout button
- Add header UI for Agenda screen (Date Picker placeholder, User Acronym)
- Add helper functions for UI
- Fixed some UI hiccups on the Login and Register screens
- Adjusted the Typography
- Added 2 more units to the size modifiers

## Question 1:
Do we need to save process death on splash screen?

## Question 2:
Which is the recommended way to add the Authorization Header to the retrofit instance?
I have 2 approaches as options, one gets the AuthToken from the AuthDao but requires that dependency. The other gets the AuthToken from the IAuthApi interface companion object.

```
// .../di/AppModule.kt
@Provides
@Singleton
fun provideTaskyApi(
    converterFactory: Converter.Factory,
    @AuthDaoProdUsingBinds authDao: IAuthDao, // for Option 1 only
): TaskyApi {
    val addHeadersInterceptor = Interceptor { chain ->

        // Option 1 - uses the `AuthDao` to get the `AuthToken` from the `DataStore`.
        //          - The `AuthDao` must be used as a dependency in this option.
        //          - Must use a `runBlocking` due to the `authDao.getAuthToken()` being a suspend function.
        runBlocking(Dispatchers.IO) {
            val authToken = authDao.getAuthToken() // <-- gets AuthToken from the AuthDaoImpl
            val request = chain.request().newBuilder()
                .addHeader("x-api-key", API_KEY)
                .addHeader(
                    "Authorization", createAuthorizationHeader(authToken ?: "NULL_AUTH_TOKEN")
                )
                .build()

            chain.proceed(request)
        }

        // Option 2 - uses the `AuthToken` stored in the `companion object` of the `IAuthApi` interface.
        //          - no need for the `AuthDao` to be a dependency of this object.
        //          - no need for the `runBlocking` since the `IAuthApi.authorizationHeader` is a simple string. (Faster access?)
        val request = chain.request().newBuilder()
            .addHeader("x-api-key", API_KEY)
            .addHeader("Authorization", IAuthApi.authorizationHeader ?: "NULL_AUTH_TOKEN") // <-- direct access here
            .build()
        chain.proceed(request)
    }

    // ... rest of the function
}
```

### DIFF
--- a/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/common/util/getUserAcronym.kt
+++ b/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/common/util/getUserAcronym.kt
@@ -1,0 +1,16 @@
+package com.realityexpander.tasky.agenda_feature.presentation.common.util
+
+fun getUserAcronym(username: String): String {
+    if(username.isBlank()) return "??"
+    if(username.length<2) return username.uppercase()
+
+    println("username: $username")
+
+    val words = username.split(" ")
+    println("words: $words, size: ${words.size}")
+    if (words.size > 1) {
+        return (words[0].substring(0, 1) + words[1].substring(0, 1)).uppercase()
+    }
+
+    return username.substring(0, 2).uppercase()
+}
--- a/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/common/util/getUserAcronym.kt
+++ b/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/common/util/getUserAcronym.kt
@@ -4,13 +4,23 @@ fun getUserAcronym(username: String): String {
     if(username.isBlank()) return "??"
     if(username.length<2) return username.uppercase()
 
-    println("username: $username")
-
     val words = username.split(" ")
-    println("words: $words, size: ${words.size}")
     if (words.size > 1) {
-        return (words[0].substring(0, 1) + words[1].substring(0, 1)).uppercase()
+        return (words.first().substring(0, 1) + words.last().substring(0, 1)).uppercase()
     }
 
     return username.substring(0, 2).uppercase()
+}
+
+fun main() {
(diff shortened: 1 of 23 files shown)


app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/common/util/getUserAcronym.kt
```diff
@@ -0,0 +1,16 @@
...
+    if(username.isBlank()) return "??"
+    if(username.length<2) return username.uppercase()
+
+    println("username: $username")
+
+    val words = username.split(" ")
+    println("words: $words, size: ${words.size}")
+    if (words.size > 1) {
+        return (words[0].substring(0, 1) + words[1].substring(0, 1)).uppercase()
+    }
```
philipplackner, line 13: This logic doesn't fully seem correct. What if the user has a middle name? Then it should take the initials from the first and last name. 
Also, what if a name would accidentally have 2 spaces between first and last name, I assume the app would crash here 🤔

philipplackner (approved):
1. Already answered in a previous PR. Handling process death is not a general best practice, but rather something to keep in mind. If your screen would behave totally normal if the whole state was reset, there's no need to handle it. Also it wouldn't really happen during the splash screen.
2. What about option 3 and reading it from preferences? I find saving that in a DB table a bit overkill since there's only one row needed and that makes it hard to encrypt (unlike `EncryptedSharedPreferences` for example).

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/common/util/getUserAcronym.kt
realityexpander: AH, ok, I will update it to get the First and Last words if there are more than 1 word, not just the second word.

PR #11, opened 2022-11-04
0009 dropdown menu handling

realityexpander:
- Built more UI elements for Agenda Screen
- Add Task, Reminder and Event basic data structures
- Make drop-down menu location based on screen location
- Add serialization and `Parcelable` for `LocalDateTime` to survive process death
- Add real day names, month name, and day of month
- Add ability to complete task
- Add FAB
- Add menu handling for different AgendaItem types

## Question 1
  - Removing `.verticalScroll()` only shows single composable and does not display lower composables.
  - How to make lower composables visible without using `.verticalScroll()`?
  
https://github.com/realityexpander/Tasky/blob/4b938dbc18c1cd728c51c9866b76ec98e16fa199/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaScreen.kt#L229


app/build.gradle
```diff
@@ -141,6 +141,7 @@ dependencies {
     // KotlinX Serialization
     implementation "org.jetbrains.kotlinx:kotlinx-serialization-json:$serialization_version"
     implementation "com.jakewharton.retrofit:retrofit2-kotlinx-serialization-converter:0.8.0"
+    implementation "org.jetbrains.kotlinx:kotlinx-datetime:0.4.0"
```
philipplackner, line 144: Any reason for not choosing the Java DateTime API which offers quite more functionality?

app/src/main/java/com/realityexpander/tasky/agenda_feature/domain/AgendaItem.kt
```diff
@@ -1,32 +1,151 @@
...
+    override val descriptor = PrimitiveSerialDescriptor("LocalDateTime", PrimitiveKind.STRING)
+
+    override fun deserialize(decoder: Decoder): LocalDateTime {
+        return LocalDateTime.parse((decoder.decodeString()))
+    }
+
+    override fun serialize(encoder: Encoder, value: LocalDateTime) {
+        encoder.encodeString(value.toString())
+    }
+}
```
philipplackner, line 32: This is against the purpose of domain, since it reveals implementation details (what kind of JSON/serialization lib you use). Why not use Parcelable + `@Parcelize`?

app/src/main/java/com/realityexpander/tasky/agenda_feature/domain/AgendaItem.kt
```diff
@@ -1,32 +1,151 @@
...
+        ) = Event(
+            id = id,
+            title = title,
+            description = description,
+            remindAt = remindAt,
+            from = from,
+            to = to,
+            attendeeIds = attendeeIds,
+            photos = photos,
+        )
```
philipplackner, line 82: I'd actually prefer to just make `Event`, `Task` and `Reminder` all data classes here, so you don't need to create copy, equals and hashCode functions for them which are all really helpful. You just need to use different field names for them than for the `AgendaItem` constructor

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
```diff
@@ -122,18 +214,32 @@ class AgendaViewModel @Inject constructor(
...
                     )
                 }
+                yield()
+            }
+            is AgendaEvent.ToggleTaskCompleted -> {
+                _agendaState.update {
+                    it.copy(
+                        agendaItems = it.agendaItems.map { agendaItem ->
+                            if (agendaItem.id == event.agendaItemId) {
+                                (agendaItem as AgendaItem.Task).copy(isDone = !agendaItem.isDone)
```
philipplackner, line 230: Maybe add an additional safety-check here for `agendaItem is AgendaItem.Task` just to avoid this accidentally gets called for a non-Task

philipplackner (approved):
I don't fully understand the problem with `verticalScroll`, but I'm wondering why you don't use a `LazyColumn` for this? It's in the end a list that could be quite large

app/src/main/java/com/realityexpander/tasky/agenda_feature/domain/AgendaItem.kt
realityexpander: I have to use both `@Parcelize` and `@Serialize`. The `SavedStateHandle` uses `@Parcelize` under the hood, and the navigation library im using uses `@Serialize`. 

Where should I move the LocalDateTimeSerializer to? Which package?

app/src/main/java/com/realityexpander/tasky/agenda_feature/domain/AgendaItem.kt
realityexpander: OK, I will separate them out.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
realityexpander: OK I will add this.

app/build.gradle
realityexpander: I dont know the difference actually!. I will look at the Java DateTime Api.

realityexpander:
> I don't fully understand the problem with `verticalScroll`, but I'm wondering why you don't use a `LazyColumn` for this? It's in the end a list that could be quite large

OK, i will implement a `LazyColumn` and see if this resolves the issue.

app/build.gradle
realityexpander: I have implemented the Java DateTime Api, it seems like a drop-in replacement.

PR #12, opened 2022-11-05
0010 add agenda item menu

realityexpander:
- Fixup Acronym UI item to not use Canvas
- Refactor `AgendaItem` to use independent `data class`es
- Refactor to use Java `DateTime` class
- Use `LazyList` for `AgendaItems`
- Improve color & font handling in menus
- Add FAB menu functionality
- Add ability of `MenuItem` to use either `Painter` or `ImageVector` for icon
- Add "scroll to item" functionality after an item is created.
- Added one-time events to the `AgendaState` and mechanism to reset it after a task is finished. 

### About my one-time-event mechanism:

 I tried using `SharedFlow`, but because it is a Hot Flow, I was getting a race conditions.

In this case, I was attempting to scroll to the bottom of the list after an `AgendItem` is added. I was using a `LaunchedEffect` and making the changes based on when the `AgendaItems` list was updated. 

This caused a race condition that re-launched the `LaunchEffect` and in the middle of the scroll (or before it started), the coroutine was cancelled and the `LaunchEffect` recomposed with the latest data from the `AgendaItems`.

Maybe there is a better way to do this, but I like using the `StateFlow` to synchronize the state so I can be sure the order of things, ie: 

1. Create `AgendaItem` ->
2. `AgendaItem` list is updated ->
3. `ScrollTo` the new item `one-time event` initiated ->
4. Scroll happens in Compose coroutine scope ->
5. Scroll completed ->
6. one-time event cleared. (finished sequence)

Relevant code :
https://github.com/realityexpander/Tasky/blob/f4c77442b1f61aa4f83e535b94ec20fd39754f0c/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt#L223

https://github.com/realityexpander/Tasky/blob/f4c77442b1f61aa4f83e535b94ec20fd39754f0c/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt#L295

This is probably bad somehow 😃, but i remember hearing you say in a one of your recent videos about how you would want to use `StateFlow` for all events (no channels or SharedFlow) and this was my first stab at it with the use-case of having to synchronize events and prevent coroutine cancellations due to race conditions that were caused by using a hot flow emitter.

Let me know what you think!

## Question 1 
DropDown menu is not able to be "offset" to the very bottom of the screen. It always has about 75 dp of "padding" at the bottom of the menu. (see pic)

I looked around online and could not find an answer for this. Is this normal behavior? Is it possible to have the drop-down on the very bottom of the screen? My dropdowns are all inside a `Box` composable and otherwise work as expected.

image

### DIFF
--- a/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaEvent.kt
+++ b/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaEvent.kt
@@ -1,5 +1,6 @@
 package com.realityexpander.tasky.agenda_feature.presentation.agenda_screen
 
+import com.realityexpander.tasky.agenda_feature.presentation.common.enums.AgendaItemType
 import com.realityexpander.tasky.core.presentation.common.util.UiText
 import com.realityexpander.tasky.core.util.UuidStr
 
@@ -7,27 +8,22 @@ sealed interface AgendaEvent {
     data class SetIsLoaded(val isLoaded: Boolean) : AgendaEvent
     data class ShowProgressIndicator(val isShowing: Boolean) : AgendaEvent
 
-    object ValidateUsername : AgendaEvent
-
-    object ShowInvalidUsernameMessage : AgendaEvent
-
-    data class UnknownError(val message: UiText) : AgendaEvent
-
-    data class CreateEvent(
-            val username: String,
-            val email: String,
-            val password: String,
-            val confirmPassword: String
-        ) : AgendaEvent
-    data class CreateEventSuccess(val message: UiText) : AgendaEvent
-    data class CreateEventError(val message: UiText) : AgendaEvent
-
-
+    // • Logout
     object ToggleLogoutDropdown : AgendaEvent
     object Logout : AgendaEvent
 
-    data class ShowAgendaItemDropdown(val agendaItemId: UuidStr? = null) : AgendaEvent
+    // • Agenda Item - Create
+    object ToggleCreateAgendaItemDropdown : AgendaEvent
+    data class CreateAgendaItem(val agendaItemType: AgendaItemType) : AgendaEvent
+    data class CreateAgendaItemSuccess(val message: UiText) : AgendaEvent
+    data class CreateAgendaItemError(val message: UiText) : AgendaEvent
 
+    // • Agenda Item - Open/Edit/Delete
+    data class ShowAgendaItemActionDropdown(val agendaItemId: UuidStr? = null) : AgendaEvent
+
+    // • Task Item - Quick-edit
     data class ToggleTaskCompleted(val agendaItemId: UuidStr) : AgendaEvent
 
+    // • Errors
+    data class UnknownError(val message: UiText) : AgendaEvent
 }
--- a/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaEvent.kt
+++ b/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaEvent.kt
@@ -25,9 +25,9 @@ sealed interface AgendaEvent {
     // • One-time events
     sealed interface StatefulOneTimeEvent {
         object ResetScrollTo                                        : StatefulOneTimeEvent, AgendaEvent
-        data class ScrollToItemId(val agendaItemId: UuidStr)  : StatefulOneTimeEvent, AgendaEvent
         object ScrollToTop                                          : StatefulOneTimeEvent, AgendaEvent
         object ScrollToBottom                                       : StatefulOneTimeEvent, AgendaEvent
+        data class ScrollToItemId(val agendaItemId: UuidStr)        : StatefulOneTimeEvent, AgendaEvent
     }
 }
 
--- a/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaEvent.kt
+++ b/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaEvent.kt
@@ -29,8 +29,10 @@ sealed interface AgendaEvent {
 
     // • One-time events
     sealed interface OneTimeEvent {
-        object ResetOneTimeEvent : OneTimeEvent, AgendaEvent
+        object Reset : OneTimeEvent, AgendaEvent
         data class ScrollToItem( val agendaItemId: UuidStr) : OneTimeEvent, AgendaEvent
+        object ScrollToTop : OneTimeEvent, AgendaEvent
+        object ScrollToBottom : OneTimeEvent, AgendaEvent
     }
 }
 
--- a/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaEvent.kt
+++ b/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaEvent.kt
@@ -27,9 +27,10 @@ sealed interface AgendaEvent {
     // • Errors
     data class UnknownError(val message: UiText) : AgendaEvent
 
-    // • Scroll to item
-    data class ScrollToItem(val itemId: UuidStr) : AgendaEvent
-
-    // • Null event
-    object Null : AgendaEvent
+    // • One-time events
+    sealed interface OneTimeEvent {
+        object ResetOneTimeEvent : OneTimeEvent, AgendaEvent
+        data class ScrollToItem( val agendaItemId: UuidStr) : OneTimeEvent, AgendaEvent
+    }
 }
+
--- a/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaEvent.kt
+++ b/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaEvent.kt
@@ -8,21 +8,14 @@ sealed interface AgendaEvent {
     data class SetIsLoaded(val isLoaded: Boolean) : AgendaEvent
     data class ShowProgressIndicator(val isShowing: Boolean) : AgendaEvent
 
-    // • Logout
-    object ToggleLogoutDropdown : AgendaEvent
     object Logout : AgendaEvent
 
     // • Agenda Item - Create
-    object ToggleCreateAgendaItemDropdown : AgendaEvent
     data class CreateAgendaItem(val agendaItemType: AgendaItemType) : AgendaEvent
     data class CreateAgendaItemSuccess(val message: UiText) : AgendaEvent
     data class CreateAgendaItemError(val message: UiText) : AgendaEvent
 
-    // • Agenda Item - Open/Edit/Delete
-    data class ShowAgendaItemActionDropdown(val agendaItemId: UuidStr? = null) : AgendaEvent
-
-    // • Task Item - Quick-edit
-    data class ToggleTaskCompleted(val agendaItemId: UuidStr) : AgendaEvent
+    data class TaskToggleCompleted(val agendaItemId: UuidStr) : AgendaEvent
 
     // • Errors
     data class UnknownError(val message: UiText) : AgendaEvent
--- a/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaEvent.kt
+++ b/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaEvent.kt
@@ -23,11 +23,11 @@ sealed interface AgendaEvent {
     data class Error(val message: UiText) : AgendaEvent
 
     // • One-time events
-    sealed interface OneTimeEvent {
-        object Reset : OneTimeEvent, AgendaEvent
-        data class ScrollToItem( val agendaItemId: UuidStr) : OneTimeEvent, AgendaEvent
-        object ScrollToTop : OneTimeEvent, AgendaEvent
-        object ScrollToBottom : OneTimeEvent, AgendaEvent
+    sealed interface StatefulOneTimeEvent {
+        object Reset : StatefulOneTimeEvent, AgendaEvent
+        data class ScrollToItem( val agendaItemId: UuidStr) : StatefulOneTimeEvent, AgendaEvent
+        object ScrollToTop : StatefulOneTimeEvent, AgendaEvent
+        object ScrollToBottom : StatefulOneTimeEvent, AgendaEvent
     }
 }
 
--- a/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaEvent.kt
+++ b/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaEvent.kt
@@ -26,4 +26,10 @@ sealed interface AgendaEvent {
 
     // • Errors
     data class UnknownError(val message: UiText) : AgendaEvent
+
+    // • Scroll to item
+    data class ScrollToItem(val itemId: UuidStr) : AgendaEvent
+
+    // • Null event
+    object Null : AgendaEvent
 }
--- a/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaEvent.kt
+++ b/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaEvent.kt
@@ -10,6 +10,8 @@ sealed interface AgendaEvent {
 
     object Logout : AgendaEvent
 
+    data class SetSelectedDayIndex(val dayIndex: Int) : AgendaEvent
+
     // • Agenda Item - Create
     data class CreateAgendaItem(val agendaItemType: AgendaItemType) : AgendaEvent
     data class CreateAgendaItemSuccess(val message: UiText) : AgendaEvent
--- a/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaEvent.kt
+++ b/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaEvent.kt
@@ -24,10 +24,10 @@ sealed interface AgendaEvent {
 
     // • One-time events
     sealed interface StatefulOneTimeEvent {
-        object Reset : StatefulOneTimeEvent, AgendaEvent
-        data class ScrollToItem( val agendaItemId: UuidStr) : StatefulOneTimeEvent, AgendaEvent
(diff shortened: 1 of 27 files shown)


app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaScreen.kt
```diff
@@ -146,6 +155,47 @@ fun AgendaScreenContent(
...
+                        scope.launch {
+                            scrollState.animateScrollToItem(item)
+                            onAction(AgendaEvent.OneTimeEvent.Reset)
+                        }
+                    }
+                }
+                else -> {}
+            }
+        }
+    }
```
philipplackner, line 174: Not a good idea, since it's very vulnerable to race conditions. If event A hasn't been processed yet and event B is assigned, this whole `LaunchedEffect` block will be cancelled and relaunched. 
Also, you currently don't have a way to reset the one time event. Imagine you show an error here and when a config change happens, this would be re-triggered and show the error again, so there's not really anything one time about this.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaScreen.kt
```diff
@@ -146,6 +155,47 @@ fun AgendaScreenContent(
...
+//                        actionLabel = oneTimeEvent.actionLabel,
+//                        duration = SnackbarDuration.Short,
+//                        onDismiss = {
+//                            onAction(AgendaEvent.OneTimeEvent.ResetOneTimeEvent)
+//                        },
+//                        onActionClick = {
+//                            onAction(AgendaEvent.OneTimeEvent.ResetOneTimeEvent)
+//                            oneTimeEvent.onActionClick()
+//                        }
+//                    )
```
philipplackner, line 192: Especially here where you have a suspending function in `LaunchedEffect`, the code in this would be active as long as the snackbar is visible. As soon as another one time event is assigned, the snackbar would be hidden because the event was overridden.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaScreen.kt
```diff
@@ -510,17 +554,61 @@ fun AgendaScreenContent(
...
+            offset = DpOffset(
+                x = fabButtonOffset.x.dp,
+                y = -screenHeight + fabButtonOffset.y.dp
+            ),
+            properties = PopupProperties(
+                usePlatformDefaultWidth = true,
+            ),
+            modifier = Modifier
+                .background(color = MaterialTheme.colors.onSurface)
+        ) {
```
philipplackner, line 571: Little tip: You can put your `DropdownMenu` in a box with the view it should be anchored to (like the FAB here) and it'll automatically open above/below it. For example, this is my dropdown for logging out:
```
Box {
    DropdownMenu(
        expanded = state.isSelectingUserOptions,
        onDismissRequest = {
            viewModel.onToggleUserOptions(false)
        }
    ) {
        DropdownMenuItem(onClick = viewModel::logout) {
            Text(text = stringResource(id = R.string.logout))
        }
    }
    UserAvatar(
        avatarInitials = state.userInitials,
        modifier = Modifier
            .size(40.dp),
        onAvatarClick = {
            viewModel.onToggleUserOptions(true)
        }
    )
}
```

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaScreen.kt
```diff
@@ -646,7 +734,49 @@ fun AgendaScreenPreview_NightMode_NO() {
...
+//                            ) {
+//                                Text(
+//                                    text = stringResource(id = R.string.cancel),
+//                                    fontWeight = FontWeight.Bold,
+//                                )
+//                            }
+//                        },
+//                    )
+//                }
+//            }
```
philipplackner, line 779: I'm personally not a friend of including commented out code in commits, since that doesn't add any value and just makes them larger. Deleting this code is fine - if you'll ever need it again Git got your back.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaState.kt
```diff
@@ -17,9 +17,17 @@ data class AgendaState(
...
     val isLogoutDropdownVisible: Boolean = false,
+
+    // Menu for editing/viewing/deleting agenda items
     val agendaItemIdForMenu: UuidStr? = null,
 
+    // Menu for adding agenda items
+    val isAddAgendaItemDropdownVisible: Boolean = false,
+
+    // One-time events
+    val oneTimeEvent: AgendaEvent? = null,
```
philipplackner, line 30: I also find the name `oneTimeEvent` misleading, since it's modeled as state which is everything but one time

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
```diff
@@ -251,6 +292,25 @@ class AgendaViewModel @Inject constructor(
...
+                        }
+                        yield()
+                    }
+                    else -> {
+                        // Send the one time event
+                        yield()
+                        _agendaState.update {
+                            it.copy(oneTimeEvent = event)
+                        }
+                        yield()
```
philipplackner, line 310: I'm actually curious why you use all these `yield()` calls across your project, I have never used that in my life and was always good 🤔 😂 
What would break if you would remove these calls?

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/components/MenuItem.kt
```diff
@@ -50,18 +52,33 @@ fun MenuItem(
...
+                    tint = if (enabled)
+                        MaterialTheme.colors.surface
+                    else
+                        MaterialTheme.colors.surface.copy(alpha = 0.2f),
+                    modifier = Modifier
+                        .padding(DP.tiny)
+                        .align(Alignment.CenterVertically)
+                )
+            }
         }
```
philipplackner, line 82: The outer if-check seems unnecessary to me

philipplackner (approved):
1. Regarding your one time event logic, my comments should explain why your implementation can be problematic. Here is an example of how I handle showing snackbar messages for example:
```
// In TaskDetailViewModel
...
is AuthResult.Error -> {
    updateState {
        it.copy(
            infoMessage = UiText.unknownError(),
            isLoadingTask = false
        )
    }
}
...
fun onInfoMessageSeen() {
    updateState { it.copy(infoMessage = null) }
}
...

// In TaskDetailScreen.kt
LaunchedEffect(
    key1 = state.infoMessage,
    key2 = scaffoldState.snackbarHostState
) {
    state.infoMessage?.let { message ->
        scaffoldState.snackbarHostState.showSnackbar(
            message = message.asString(context)
        )
        viewModel.onInfoMessageSeen()
    }
}
``` 
3. Not sure why you want this, I think it's better UX if the user sees that clicking on the FAB actually expands the dropdown (so that it shows up above it). Haven't had that situation yet, but as I explained in a comment (and it seems you already know that), you can put the dropdown in a `Box` to use as an anchor. Have you tried putting it in a box together with a no-height spacer that's below the FAB?

realityexpander:
> 3. Not sure why you want this, I think it's better UX if the user sees that clicking on the FAB actually expands the dropdown (so that it shows up above it). Haven't had that situation yet, but as I explained in a comment (and it seems you already know that), you can put the dropdown in a `Box` to use as an anchor. Have you tried putting it in a box together with a no-height spacer that's below the FAB?

I am wanting the Drop-down to show up over the FAB, just like in the other parts of the screen where the "skewer (three dots)" menu drop-downs appear over the "skewers".

I currently have the drop-down menu in a `Box` with zero padding or spacers.  Unlike other components, there is always a "lower area" that the drop-down cannot be `Offset` into. I can put other composables in there (like buttons, text, etc) but the drop-down will not go lower than about 75 dp from the bottom. 

Also not sure what you mean by "anchor". The dropdown is in a Box (with the other menus and FAB). I can put Text or Buttons wherever I like within this `Box`. 

I have been using the `.align(alignment = Alignment.BottomEnd)` and it is ignored by the `DropDownMenu`, even when I don't use `offset = xxx` parameter.

This seems to be a limitation of the `DropdownMenu` implementation, and if so, I will leave as such. Just wondering if I was crazy or that it was a known issue. I see nothing online about it.

image

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaEvent.kt
```diff
@@ -1,33 +1,38 @@
...
+    data class CreateAgendaItemError(val message: UiText) : AgendaEvent
+
+    // • Agenda Item - Open/Edit/Delete
+    data class ShowAgendaItemActionDropdown(val agendaItemId: UuidStr? = null) : AgendaEvent
 
+    // • Task Item - Quick-edit
     data class ToggleTaskCompleted(val agendaItemId: UuidStr) : AgendaEvent
 
+    // • Errors
+    data class UnknownError(val message: UiText) : AgendaEvent
```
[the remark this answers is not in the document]
realityexpander: OK I will change this be just "error"

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaScreen.kt
```diff
@@ -89,17 +95,20 @@ fun AgendaScreenContent(
...
+    // For positioning menus
     var logoutButtonOffset by remember { mutableStateOf(Offset.Zero)}
-
-    val agendaItems = state.agendaItems
+    var fabButtonOffset by remember { mutableStateOf(Offset.Zero)}
 
     // Keeps track of the dropdown menu offsets
     val agendaItemMenuInfos = remember { mutableStateMapOf<UuidStr, MenuItemInfo>() }
 
+    var selectedDay by remember { mutableStateOf(0) }
```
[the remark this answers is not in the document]
realityexpander: OK, i will make this part of the AgendaState.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaScreen.kt
realityexpander: On Line 167, I am sending a "reset" `OneTimeEvent.Reset` here, only after the "scrolling" is completed.  I have an `onEvent` handler in `AgendaViewModel` to reset the one-time event. Maybe you missed that?

The whole reason I do it this way is to prevent a race condition that was happening when i attempted to use a one-time event using a SharedFlow. 

What I'm trying to do is to scroll to the new item AFTER it has been added to the list. And since the block inside the launched effect is captured as a closure, I dont have access to the current value of `agendaItems` until AFTER it gets updated. And the race condition was occurring when the `agendaItems` is updated when i was using a `SharedFlow` and the `ScrollTo` event was lost because it was cancelled. 

I want to make sure I only trigger the `scrollTo` AFTER the `agendaItems` has been updated. It has to be sequential in this case. This particular one-time event is depending on the data being available from the `agendaState` `stateIn` in the ViewModel. The `ScrollTo` needs to have the current `AgendaItems` available to be able to do a search for the `agendaItem` id. Due to this, only after the data in `agendaItems` has been updated can I perform the "ScrollTo".

So this is not like showing a "snackbar" or "errorMessage". I will have a separate one-time event handler for that using `Channels` or `SharedFlow`. 

I hope this make sense.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaScreen.kt
realityexpander: This code is commented out and only a placeholder. The implementation is incorrect and has not been written yet. 

I am trying to make the PR's smaller, so I am not implementing all the functionality at once, and leaving placeholders for this. This is why it is commented out.

I will be more explicit in the future about commented-out code being used as a placeholder.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaScreen.kt
realityexpander: OK, i will try out this implementation. It should save a lot of confusing code!

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaScreen.kt
realityexpander: OK, i will remove code like this from future PR's

realityexpander: Is there an easy way to go back to this code, or do you have to look through each PR and Commit? (thats the only way I know how to do it, other than stashing)...

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaState.kt
realityexpander: I have it as a stateful one-time event. I have a reset, and it similar to your pattern of a `ShowXXXXEvent` and `SeenXXXEvent`, just more generalized.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
realityexpander: I want to be sure the "AgendaItems" are updated properly, otherwise I get race conditions. 

Is there another way to update `AgendaItems` and have it processed by the `stateIn` fully before proceeding to the next step? Otherwise the "intermediate" states could be lost or ignored. The `stateIn` function needs time to run on the "looper". Without using the `yield()`, I was getting race conditions.

Can you suggest another way of structuring my code in a way that will respect the `stateIn` processing and allow events to be seen and not ignored?

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/components/MenuItem.kt
realityexpander: Not sure what you mean here. You mean the check for `painterIcon` or `vectorIcon`? The `Icon` composable only allows for one or the other. You cant have both. The compiler gives an error.

realityexpander:
I am having a problem with the "Compose-Destinations" library where it does not give the `SaveStateHandle`, so I have a temporary hack workaround (that actually works very good). I have contacted the developer of the library for help, but if you see something glaringly obvious that Im doing wrong here, i would be appreciative for any hints.

The savedStateHandle works find without the library, so Im very confused about whats going on.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaScreen.kt
philipplackner: Ah I see regarding the reset mechanism, I missed that. In your case it's still a problem since the reset action is fired AFTER the scroll animation which means if during the scroll another event comes in, the reset action of the scroll will reset the new event. 
It also still comes with the problem that there can be only one one-time event at a time which is quite against the concept of concurrency. If you really want to use these one-time events I'd just stick to Channels and collect them in `LaunchedEffect(Unit)`. 
If you want to do with with states, you need to make each one-time event a separate state so you allow that there can be multiple at a time. 

Regarding your issue (hope I get you right): Why not do something like this which is executed every time `agendaItems` changes. The logic for `shouldScrollToItem` could be a state or a function of your ViewModel:
```
LaunchedEffect(state.agendaItems) {
    if(shouldScrollToItem) {
        // Scroll to item
    }
}
```

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaScreen.kt
philipplackner: I personally just use the Git commit history in Android Studio and go back to a commit where I know this code was contained in. You can then just right click and compare with local where you can easily include this code with a click on that little diff arrow in the commit-difference tab.

Works well if you work alone on the project. If others are also working on it, saving snippets sounds like a good approach, since then code gets easier lost and you could also better share these snippets with co-workers (since you might not even know there is a snippet)

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaState.kt
philipplackner: Wrote a bit about that in my first comment, but how do you know there can only be one one-time event at a time? Couldn't it be that a snackbar shows while you perform a scroll to the latest item?

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
philipplackner: Alright, usually `stateFlow.update` should take care of all race conditions, but if it works for you I'm not saying anything against that

philipplackner:
Not sure what exactly the issue is, but I'm also using that lib in my project and it works fine. Here the versions I'm using:
Compose compiler:
`compose_compiler_version = '1.3.0-rc01'`

KSP:
`id 'com.google.devtools.ksp' version '1.7.10-1.0.6'`

Kotlin:
`id 'org.jetbrains.kotlin.android' version '1.7.10' apply false`

Compose destinations:
```
implementation 'io.github.raamcosta.compose-destinations:core:1.6.13-beta'
ksp 'io.github.raamcosta.compose-destinations:ksp:1.6.13-beta'
```

By passing the parameter to the `@Destination` annotated screen, it will be included in SavedStateHandle (I think there's also a way without passing it, but don't know that).

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaScreen.kt
realityexpander: OK, I see your code example is a better way to do it. Let me implement that solution and check it out.

PR #13, opened 2022-11-08
0011 add event repo

realityexpander:
* Added Agenda Repository
* Added Event Repository
* Added Mapper functions (DTO <-> Domain <-> Entity)
* Added fake database and api implementation for Agenda and Event repository
* Changed to ZonedDateTime
* Added converters from/to UTC for ZonedDateTime
* Removed non-domain code from AgendaItems
* Created starter Tasky database with Event entity
* Created TypeConverters for ZonedDateTime and LocalDateTime
* Created TypeParcelers for ZonedDateTime and LocalDateTime
* Created Serializer for ZonedDateTime
* Added logout call to TaskyApi
* Flushed out Event data class with all properties
* Added AgendaSync data classes (DTO & Domain)
* Added stub Room database (not implemented yet)
* `Task` and `Reminder` are not implemented yet, I want to be sure Im creating a good pattern first with `Event`

Sorry this PR is a bit large, but its mostly boilerplate code, and not yet fully functional. I want to confirm that im heading in the right direction with this before going further.

## Question 1

Is the package setup correct? I have each Agenda Item Repository setup with its own DAO and API, mirroring the "by-feature" concept instead of "by layer"

Im still a little confused where to put things, like common `TypeParcelers` and `serializers` and `typeAliases`

## Question 2

I also wanted to confirm that my setup in the `AgendaRepositoryImpl` is a good direction. The `AgendaRepository` doesnt have its own table (As far as I can see so far) but uses the data from the tables of the other Entities, like `Event`, `Task` and `Reminder`.

Im also planning on keeping the `deleted` and `uploaded` flags inside each Entity, and using them as the SSOT for syncing.

Ive used the WorkManager before and watched your videos on it, but Im still a bit unsure about how much access to the database I will have. Im assuming its able to access all these different DAO's and API's with no problems.

Anyway, thanks again for this, I am learning a tremendous amount and definitely feel like I'm leveling up my skills and understand about how to structure medium-to-large size projects. 😄 

## Question 3

In the `EventConverters.kt`, I am converting from `UTC` to `ZonedDateTime` when I convert the DTO to/from the Domain. Is this a good idea? 

It seems like the right place to do it, since its the boundary between the API and the Domain. And the API uses `UTC` `Long`s to represent dates and times, and I will be using `ZonedDateTime` in the UI & Rest of app.


app/src/main/java/com/realityexpander/tasky/agenda_feature/data/common/convertersDTOEntityDomain/EventConverters.kt
```diff
@@ -0,0 +1,87 @@
...
+            to = it.to,
+            remindAt = it.remindAt,
+            host = it.host,
+            isUserEventCreator = it.isUserEventCreator,
+            isGoing = it.isGoing,
+            attendeeIds = it.attendeeIds,
+            photos = it.photos,
+            deletedPhotoKeys = it.deletedPhotoKeys,
+        )
+    }
```
philipplackner, line 26: Curious why you explicitly call `let` so you always have to prepend `it` here? 🤔

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
```diff
@@ -0,0 +1,67 @@
...
+//                    deleteReminderIds = deletedReminderIds,                       // todo implement reminders repo
+                ).toDTO()
+            )
+
+        if (deletedSuccessfully) {
+           return eventRepository.deleteFinallyEventIds(deletedEventIds)
+        }
+
+        return false
+    }
```
philipplackner, line 46: For this function it's very unclear to me what the return boolean represents. Could that maybe be reflected by either renaming the function or the return value (in a sealed interface maybe)?

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/remote/AgendaApiImpl.kt
```diff
@@ -0,0 +1,42 @@
...
+            if (response.isSuccessful) {
+                val responseBody = response.body()
+
+                responseBody ?: throw Exception("Response body is null")
+            } else {
+                throw Exception("Error getting agenda")
+            }
+        } catch (e: Exception) {
+            throw e
+        }
```
philipplackner, line 30: If the catch block stays like that, the whole try/catch could be omitted

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/remote/AgendaApiImpl.kt
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    override suspend fun syncAgenda(agendaSync: AgendaSyncDTO): Boolean {
+        return try {
+            val response = taskyApi.syncAgenda(agendaSync)
+
+            response.isSuccessful
+        } catch (e: Exception) {
+            throw e
+        }
```
philipplackner, line 40: Same here

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/eventRepository/eventRepositoryImpls/EventRepositoryImpl.kt
```diff
@@ -0,0 +1,89 @@
...
+) : IEventRepository {
+
+    override suspend fun createEvent(event: AgendaItem.Event): Boolean {
+        return try {
+            eventDao.createEvent(event.toEntity())
+            eventApi.createEvent(event.toDTO())
+            true
+        } catch (e: Exception) {
+            false
+        }
```
philipplackner, line 27: Here, you're potentially catching CancellationExceptions (same in all functions below)

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/eventRepository/local/eventDao/eventDaoImpls/EventDaoFakeImpl.kt
```diff
@@ -0,0 +1,289 @@
...
+    private val eventsFlow = flow {
+        var lastEvents= emptyList<EventEntity>()
+        while(true) {
+            if(!lastEvents.containsAll(events)) {
+                lastEvents = events.toList()
+                emit(events)
+            }
+            delay(100)
+        }
+    }
```
philipplackner, line 103: You could also just use a StateFlow here, then you're immediately notified about changes

app/src/main/java/com/realityexpander/tasky/agenda_feature/util/AgendaUtil.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import com.realityexpander.tasky.core.util.UuidStr
+
+typealias EventId = UuidStr
+typealias ReminderId = UuidStr
+typealias TaskId = UuidStr
+typealias AttendeeId = UuidStr
+typealias PhotoId = UuidStr
+
+typealias TimeZoneStr = String  // ex: "America/Los_Angeles"
```
philipplackner, line 11: Cool, I like this to make the type of ID even clearer

app/src/main/java/com/realityexpander/tasky/core/domain/typeParcelers/LocalDateTimeParceler.kt
```diff
@@ -0,0 +1,16 @@
...
+object LocalDateTimeParceler : Parceler<LocalDateTime> {
+    override fun create(parcel: Parcel): LocalDateTime {
+        return LocalDateTime.parse(parcel.readString())
+    }
+
+    override fun LocalDateTime.write(parcel: Parcel, flags: Int) {
+        parcel.writeString(this.toString())
+    }
+
+}
```
philipplackner, line 16: Can't you just use the `@Parcelize` annotation instead of this?

app/src/main/java/com/realityexpander/tasky/core/domain/typeParcelers/ZonedDateTimeParceler.kt
```diff
@@ -0,0 +1,16 @@
+package com.realityexpander.tasky.core.domain.typeParcelers
+
+import android.os.Parcel
+import kotlinx.parcelize.Parceler
+import java.time.ZonedDateTime
+
+object ZonedDateTimeParceler : Parceler<ZonedDateTime> {
```
philipplackner, line 7: Same here

philipplackner (approved):
1. Looks solid overall. I personally like packaging by context instead of by type, so for the agenda feature, you could have packages `event`, `reminder`, `task`, `syncing` etc. and these would contain everything presentation/event related, domain/event related and so on. But if you package by type (like you mostly did in your project by having packages like `enums` or `repositories`) that also works as long as you use feature packages, since a feature is always limited in size. In the end what counts is that you're consistent with your packaging and you + other people would quickly find something which I'd say is the case in your project.
2. Yes, sounds good, that's exactly what repositories are used for. Simply combining multiple data sources and forwarding what needs to be forwarded. Apart from `deleted` flags, you'll also need `updated` and `created` flags in case a user updates an event without connection or creates a new one. I personally created a separate table for modified events/tasks/reminders, that looks like this:
```
@Entity
data class ModifiedEventEntity(
    @PrimaryKey(autoGenerate = false) val eventId: String,
    val type: ModificationType // could be Created, Updated or Deleted
)
```
This minimizes the extra info you save for each item in the DB, since you only save that extra info for those items that actually changed.

And yes, you can normally inject dependencies into WorkManager with Hilt like your repositories and use them as you want there. It's quite convenient.
3. Yes the mapper is the right place for that.

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/common/convertersDTOEntityDomain/EventConverters.kt
realityexpander: Just habit. I will convert it to another context function.

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
realityexpander: OK thats a good idea, i will make this change.

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/remote/AgendaApiImpl.kt
realityexpander: OK, I will improve the catches.

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/remote/AgendaApiImpl.kt
realityexpander: OK I will improve the catches.

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/eventRepository/eventRepositoryImpls/EventRepositoryImpl.kt
realityexpander: OK, I will improve the catches.

app/src/main/java/com/realityexpander/tasky/agenda_feature/util/AgendaUtil.kt
realityexpander: Thank you!

app/src/main/java/com/realityexpander/tasky/core/domain/typeParcelers/LocalDateTimeParceler.kt
realityexpander: I dont think the Parcelize will convert the custom types of `LocalDateTime` and `ZonedDateTime`, so I thought I needed to use a custom `@Parceler`. I will attempt to use it without the custom parceler and see if the `@Parcelize` does this for me.

app/src/main/java/com/realityexpander/tasky/core/domain/typeParcelers/ZonedDateTimeParceler.kt
realityexpander: I dont think the Parcelize will convert the custom types of LocalDateTime and ZonedDateTime, so I thought I needed to use a custom @Parceler. 

I will attempt to use it without the custom parceler and see if the @Parcelize does this for me.

app/src/main/java/com/realityexpander/tasky/core/domain/typeParcelers/LocalDateTimeParceler.kt
realityexpander: Turns out it does it automatically, so i will remove this.

PR #15, opened 2022-11-12
Add stub to toggle task w/ UI layer sanity checks

realityexpander:


app/src/main/java/com/realityexpander/tasky/MainActivity.kt
```diff
@@ -42,63 +44,83 @@ class MainActivity : ComponentActivity() {
     private val viewModel: MainActivityViewModel by viewModels()
 
     override fun onCreate(savedInstanceState: Bundle?) {
-        if(false) {
+        if (false) {
```
philipplackner, line 47: I assume just testing? Then this shouldn't make its way into a PR 😄

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/add_event_screen/AddEventViewModel.kt
```diff
@@ -0,0 +1,101 @@
...
+    private val _addEventState = MutableStateFlow(AddEventState())
+    val addEventState = _addEventState.onEach { state ->
+        // save state for process death
+        savedStateHandle[SavedStateConstants.SAVED_STATE_errorMessage] = state.errorMessage
+    }.stateIn(viewModelScope, SharingStarted.WhileSubscribed(5000), AddEventState())
+
+    init {
+        viewModelScope.launch {
+            // restore state after process death
+            _addEventState.update {
```
philipplackner, line 42: You don't need this `update` in `init`, since you can just change the default value of your state in `MutableStateFlow(defaultValue)`. That way you don't need the copy and update operation

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
```diff
@@ -64,6 +66,16 @@ class AgendaViewModel @Inject constructor(
...
+
+        // Collect the last state of agendaItems and save it for fast access.
+        // Is there a better way to do this? Better with .stateIn?
+        viewModelScope.launch {
+            agendaState.collect {
+                it.agendaItems.collect {
+                    lastAgendaItems = it
+                }
+            }
+        }
```
philipplackner, line 78: Not sure if I get you right here, but do you do this so you don't need to load agenda items from the DB again when the user switches dates? Is this really a performance problem the user will notice?

app/src/main/java/com/realityexpander/tasky/agenda_feature/util/ZonedDateTimeToStrings.kt
```diff
@@ -0,0 +1,131 @@
+package com.realityexpander.tasky.agenda_feature.util
```
philipplackner, line 1: What's the difference between your util package and your layer packages? The functions of this file are all presentation related, so maybe it would make more sense to put these in `agenda_feature->presentation->util` instead 🤔

philipplackner (approved):
1. Responded to your comment, not sure if that's really what you want. But I think in your case you'd add a new collector on every single emission of `agendaState` which all only get canceled when the ViewModel gets cleared. What might work here if you really want that is `flatMapLatest` which can be used to flatten 2 flows.
2. Yeah looks good, I usually just excluded packages where I stored DTOs, but if you're using Kotlin serialization you have a bit more fine-grained control since you know serialized classes have the corresponding annotation.
3. Only event creators are allowed to remove users. If the event creator clicks on delete attendee, the app should auto-switch to edit mode and the change should only be applied when clicking save.
4. I personally put them all completely to the right using `Arrangement.SpaceBetween` with the same end padding (I think 16dp for me)

app/src/main/java/com/realityexpander/tasky/MainActivity.kt
realityexpander: Yes just for testing and I was trying to limit the size of the PR, and left this in as a work-in-progress, otherwise the PR will be even bigger!

Otherwise, I will have to remove this, make the PR, then put it back in to keep going.

What is the correct way to do keep items that I am still working on active and keep the PR small?

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/add_event_screen/AddEventViewModel.kt
realityexpander: Ah ok, that makes sense. I will move this to the "initial state" of the  MutableStateFlow of the state

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
realityexpander: Im using it in the `onEvent` handler to check if the `agendaItem` is indeed a `Task` before sending it to the `AgendaRepository` here:

https://github.com/realityexpander/Tasky/blob/050b32373c3fad67a33a88af9ba621b7319ee571/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt#L213-L224

Is there a better way to get the current state of the `flow` of `AgendaItems`?

app/src/main/java/com/realityexpander/tasky/agenda_feature/util/ZonedDateTimeToStrings.kt
realityexpander: OK, that makes sense. Im still figuring out where to put things.

app/src/main/java/com/realityexpander/tasky/MainActivity.kt
philipplackner: Well ideally in a real work environment the PRs should be small enough to not need such test code (this one was still ~5x as big as the suggested max size). Here in this project it's not an issue since only you are working on it, but imagine you'd have coworkers and want to merge code like `if(false)` into the production branch. Such code has nothing lost there 😄

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
philipplackner: Still not sure if I get you right, but can't you just use the local agenda items list in your state to find the item with the ID that was clicked on and check that? Or just pass the task as a parameter to `TaskToggleCompleted`? It's still unclear to me why you need these nested flows

app/src/main/java/com/realityexpander/tasky/MainActivity.kt
realityexpander: OK, yes, i would not do that while on a team. Still working on making the PR's smaller.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
realityexpander: OK, I will try to use the local AgendaItems

PR #16, opened 2022-11-14
0014 add state to addEvent screen

realityexpander:
* Move default state initialization of non-coroutine functions to MutableStateFlow default initializer.
* Move Util folders to more appropriate locations
* Add "progress indicator" screen
* Consolidate the "From" and "To" date into a function `TimeDateRow()`
* Make EditMode more generic for any Text, Date, Time, or other type of editing
* Add SmallHeightHorizontalDivider()
* Made EditTextModal

## No questions at this time


app/src/main/java/com/realityexpander/tasky/agenda_feature/domain/Photo.kt
```diff
@@ -2,8 +2,8 @@ package com.realityexpander.tasky.agenda_feature.domain
 
 import android.net.Uri
 import android.os.Parcelable
-import com.realityexpander.tasky.agenda_feature.util.PhotoId
-import com.realityexpander.tasky.agenda_feature.util.UrlStr
+import com.realityexpander.tasky.agenda_feature.common.util.PhotoId
+import com.realityexpander.tasky.agenda_feature.common.util.UrlStr
```
philipplackner, line 6: One tip that will make all your future code reviewers happy is to put such major name refactorings into a separate PR (which doesn't necessarily need a review), since it just adds noise to this PR.
I know it's annoying to always think about what to put in which PR, but the more you work with others the more important it gets that the project is easy to follow for everyone :) 
(I was also blamed way too often for big PRs btw, but now that I regularly review code I understand why 😅)

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/components/EditTextModal.kt
```diff
@@ -0,0 +1,182 @@
...
+import androidx.compose.ui.unit.dp
+import com.realityexpander.tasky.core.presentation.common.modifiers.DP
+import com.realityexpander.tasky.core.presentation.theme.TaskyGreen
+import com.realityexpander.tasky.core.presentation.theme.TaskyTheme
+
+@OptIn(ExperimentalMaterialApi::class)
+@Composable
+fun EditTextModal(
+    title: String,
+    text: String = "",
```
philipplackner, line 33: A bit nitpicky, but I'd not use default values for parameters that make up the core functionality of a composable, since that increases the likelihood that you forget to assign them and then first need to check why

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/components/EditTextModal.kt
```diff
@@ -0,0 +1,182 @@
...
+
+    var editingText by remember { mutableStateOf(text) }
+    val focusRequester = remember { FocusRequester() }
+    val interactionSource = remember { MutableInteractionSource() }
+
+    DisposableEffect(true) {
+        focusRequester.requestFocus()
+
+        onDispose { /* do nothing */ }
+    }
```
philipplackner, line 47: Might as well be replaced with `LaunchedEffect(true)`

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventScreenEvent.kt
```diff
@@ -0,0 +1,107 @@
+package com.realityexpander.tasky.agenda_feature.presentation.event_screen
+
+import com.realityexpander.tasky.agenda_feature.domain.Attendee
+import com.realityexpander.tasky.agenda_feature.domain.Photo
+import com.realityexpander.tasky.core.presentation.common.util.UiText
+import java.time.ZonedDateTime
+
+sealed interface EventScreenEvent {
```
philipplackner, line 8: Overall, I find this a bit confusing to be honest. For example, it's not clear to me why there are so many nested sealed interfaces and if these are events sent to the VM or from the VM (for example what is the event `Attendees` used for?)

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/components/TimeDateRow.kt
```diff
@@ -0,0 +1,122 @@
...
+                    .align(Alignment.CenterVertically)
+            )
+        }
+        Icon(
+            imageVector = Icons.Filled.ChevronRight,
+            tint = if (isEditable) MaterialTheme.colors.onSurface else Color.Transparent,
+            contentDescription = "Edit Event `$title` Time",
+            modifier = Modifier
+                .weight(.2f)
+                .width(32.dp)
```
philipplackner, line 69: Won't `width()` override `weight()`?

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/components/TimeDateRow.kt
```diff
@@ -0,0 +1,122 @@
...
+                    if(isEditable) {
+                        onAction(
+                            EventScreenEvent.SetEditMode(
+                                EventScreenEvent.EditMode.FromTime(
+                                    date
+                                )
+                            )
+                        )
+                    }
+                }
```
philipplackner, line 82: `clickable` also has a parameter `enabled` which should be preferred here and set to `isEditable` since otherwise you'll also get ripple effects if it's not editable.

app/src/main/java/com/realityexpander/tasky/agenda_feature/domain/Photo.kt
realityexpander: OK, i will do that in the future.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/components/EditTextModal.kt
realityexpander: OK good point. Will address.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/components/EditTextModal.kt
realityexpander: OK, i figured that there is no difference.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventScreenEvent.kt
realityexpander: I didnt complete this fully, I hope it will make more sense once I add the rest of the editors. I am attempting to make the calls from the Composable Screen easier to work with and possibly reusable for other screens.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/components/TimeDateRow.kt
realityexpander: I will experiment with this. But it seems the weight is applied to the block and the width to the icon size.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/components/TimeDateRow.kt
realityexpander: Ah, ok, i was looking for that!

The composable interface has some inconsistencies in where these kinds of things are kept!

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/components/TimeDateRow.kt
philipplackner: Interesting, thanks

PR #17, opened 2022-11-15
0015 add event date time edit

realityexpander:
* Use `_agendaState.value.agendaItems.mapLatest` to get current `agendaItems` instead of my hack
* Add Date & Time pickers
* Add `RemindAt` drop-down
* Add ability to edit From, To & Remind At date/times
* Validate dates & times are before/after each other
* Sync `RemindAt` `DateTime` as the `fromDateTime` changes
* Had to rename vars for clarity (Event is an overloaded word!!!)
* Refactor packages (I will work to make these separate PR in the future)
* Created `RemindAtRow` as function
* Create UI Resource strings
* Make ZoneDateTime human readable string use resources

## Question 1
For Reminder and Task, the UI is nearly identical. Do you want them implemented as separate Composable screens, or could they be combined?


app/src/main/java/com/realityexpander/tasky/core/presentation/common/util/UiText.kt
```diff
@@ -245,6 +245,22 @@ open class UiText : Parcelable {
         }
     }
 
+    // Useful for plain kotlin classes that don't have access to @Composable functions.
+    // Returns null if Str is `None` or `Str` is null value. Ignores the Resource value
+    // (good for testing)
+    fun asStrOrNull(): String? {
+        return when(this) {
+            is None -> null
```
philipplackner, line 253: Why not make the `UiText` nullable and get rid of `None`?

philipplackner (approved):

app/src/main/java/com/realityexpander/tasky/core/presentation/common/util/UiText.kt
realityexpander: I will remove it

PR #18, opened 2022-11-16
0016 add remove attendee

realityexpander:
* Add ability to Add `Attendees` via email
   * Add custom Material Design dialog to add an Attendee
   * Show progress in the dialog
   * Show error messages in the dialog
* Add ability to delete `Attendees`
   * Added confirmation dialogs to delete
   * Access Api to check for valid emails
* Refactor the `RepositoryResponse` to return a generic payload and use `UiText`
* Add `AttendeeRepository` (just calls the API)
* Make dialogs use resource strings
* make error messages use resource strings (unless coming from the api)
* Added `AttendeeRepository` to Hilt

## Question 1
Are the emails saved "all lower case" on the backend? Should we be converting emails to lower case before sending them to the backed (during registration)?

## Question 2
Is this an acceptable way to handle `Success/Error` for doing calls from the `EventViewModel`? 
Instead of using `try/catch` blocks in the `EventViewModel`?

Defined here:
https://github.com/realityexpander/Tasky/blob/4b32019ed79dffcef010f6aec5b35205fd44a160/app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt#L79-L93

Used here:
https://github.com/realityexpander/Tasky/blob/6498886db6580a98dae37577eb7c518980c5adf3/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventViewModel.kt#L174-L221


app/src/main/java/com/realityexpander/tasky/agenda_feature/common/RepositoryResult.kt
```diff
@@ -1,6 +1,15 @@
...
+    data class Success<T>(
+        val data: T? = null,
+        val message: String? = null
+    ) : RepositoryResult<T>
+
+    data class Error<T>(
+        val message: UiText,
+        val originalMessage: String? = null
+    ) : RepositoryResult<T>
 }
```
philipplackner, line 15: I'd suggest calling this something a bit more generic, since something to distinguish between success and error might be used for more than just in repositories

app/src/main/java/com/realityexpander/tasky/agenda_feature/common/RepositoryResult.kt
```diff
@@ -1,6 +1,15 @@
 package com.realityexpander.tasky.agenda_feature.common
```
philipplackner, line 1: Instead of using a `common` package inside a feature, you'll almost always be able to clearly put a class in a certain package. This `RepositoryResult` will fit well into domain. If it's some kind of util function for the UI, it should go in `feature->presentation->util` for example and same for data. 
In the end, the different layers are there for a reason and putting stuff into a separate `common` package will completely separate it from any layer.

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
```diff
@@ -50,27 +52,43 @@ class AgendaRepositoryImpl @Inject constructor(
...
+    ) {
+        val result = attendeeRepository.getAttendee(attendeeEmail)
+        when(result) {
+            is RepositoryResult.Success -> {
+                onSuccess(result.data)
+            }
+            is RepositoryResult.Error -> {
+                onFailure(result.message)
+            }
+        }
```
philipplackner, line 92: Regarding your question: Sure, this works and you can do it like this. But then I wonder why you use `RepositoryResult` at all, since I don't see a reason to only use callbacks for this single function (if there is from your side, please correct me)

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/attendeeRepository/attendeeRepositoryImpls/AttendeeRepositoryImpl.kt
```diff
@@ -0,0 +1,57 @@
...
+                )
+            }
+            false -> RepositoryResult.Error(
+                UiText.Str(
+                    result.exceptionOrNull()?.message ?: "Api Unknown Error"
+                ),
+                result.exceptionOrNull()?.message
+            )
+        }
+    }
```
philipplackner, line 51: If you only have 2 branches, I always find if/else more readable than `when`

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/attendeeRepository/remote/AttendeeApiImpl.kt
```diff
@@ -0,0 +1,42 @@
...
+                response.body()?.let { attendeeExistsResponseDTO ->
+                    if (attendeeExistsResponseDTO.doesUserExist) {
+                        return Result.success(attendeeExistsResponseDTO)
+                    } else {
+                        return Result.failure(Exception("User email does not exist"))
+                    }
+                }
+            }
+
+            return Result.failure(Exception(getErrorBodyMessage(response.errorBody()?.string())))
```
philipplackner, line 28: Could it make sense to create a more specific exception here, like `NetworkException`? Otherwise, it'll be a harder to distinguish between different types of exceptions

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventScreenEvent.kt
```diff
@@ -86,11 +95,26 @@ sealed interface EventScreenEvent {
...
+        }
+        sealed interface PhotoIdPayload {
+            val photoId: PhotoId
+        }
+        sealed interface AttendeePayload {
+            val attendee: Attendee
+        }
+        sealed interface AttendeeIdPayload {
+            val attendeeId: AttendeeId
+        }
```
philipplackner, line 109: I think I asked this before and maybe I've missed the answer, but what's the benefit of wrapping these in sealed interfaces?

philipplackner (approved):
1. Yes, the backend trims them and converts them to lowercase before validating and saving them in the DB.
2. Answered in comment

app/src/main/java/com/realityexpander/tasky/agenda_feature/common/RepositoryResult.kt
realityexpander: OK, sounds good.

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/attendeeRepository/attendeeRepositoryImpls/AttendeeRepositoryImpl.kt
realityexpander: ok, i will switch back to if/then

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/attendeeRepository/remote/AttendeeApiImpl.kt
realityexpander: ok, i will add more catches for the exceptions.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventScreenEvent.kt
realityexpander: When the event is received and processed, it will have a specific type associated with it, and not a general type. Make I can associate a Generic type with it, but it allows for easy "decompose" of the type of data, like Im doing here:

https://github.com/realityexpander/Tasky/blob/134ccc3d41384e7ab4054cafaba8956e39f94fe0/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventScreen.kt#L805

realityexpander: I made a video explaining this functionality:

https://youtu.be/mqlVTJw0avY

realityexpander: https://www.loom.com/share/aa8568f1b85142a194b7aa7d65265d88

app/src/main/java/com/realityexpander/tasky/agenda_feature/common/RepositoryResult.kt
realityexpander: OK sounds good. Still a bit confused about what goes where, and starting to understand.

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
realityexpander: Good point, I will change to use the Result instead and do the checks on the ViewModel side. I think I got bad thinking habits from using callbacks for so long!

PR #19, opened 2022-11-17
0017 add validation to attendee dialog

realityexpander:
* Add validation for email
* Show attendee email validation while typing
* Clear errors in Add Attendee dialog
* Put Event Editors into separate function
* Improve event names

### DIFF
--- a/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventEditors.kt
+++ b/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventEditors.kt
@@ -1,6 +1,9 @@
 package com.realityexpander.tasky.agenda_feature.presentation.event_screen
 
-import androidx.compose.foundation.layout.*
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.text.KeyboardOptions
 import androidx.compose.material.*
 import androidx.compose.material.icons.Icons
@@ -28,7 +31,7 @@ import com.vanpra.composematerialdialogs.rememberMaterialDialogState
 import java.time.LocalDateTime
 
 @Composable
-fun Editors(
+fun EventPropertyEditors(
     editMode: EditMode,
     state: EventScreenState,
     onAction: (EventScreenEvent) -> Unit,
@@ -262,4 +265,4 @@ fun Editors(
             }
         }
     }
-}
+}n
--- a/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventEditors.kt
+++ b/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventEditors.kt
@@ -1,0 +1,265 @@
+package com.realityexpander.tasky.agenda_feature.presentation.event_screen
+
+import androidx.compose.foundation.layout.*
+import androidx.compose.foundation.text.KeyboardOptions
+import androidx.compose.material.*
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Check
+import androidx.compose.material.icons.filled.Error
+import androidx.compose.runtime.*
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.res.stringResource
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventEditors.kt
```diff
@@ -0,0 +1,268 @@
...
+            editMode as EditMode.TextPayload
+
+            EditTextModal(
+                text = editMode.text,
+                title = editMode.dialogTitle.get,
+                editTextStyle =
+                if (editMode is EditMode.ChooseTitleText)
+                    MaterialTheme.typography.h2  // cant access from non-compose function, make a wrapper?
+                else
+                    MaterialTheme.typography.body1, // cant access from non-compose function, make a wrapper?
```
philipplackner, line 51: Why does this need to be accessed from non-compose code?

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventEditors.kt
```diff
@@ -0,0 +1,268 @@
...
+                        text = stringResource(R.string.ok),
+                        disableDismiss = true,
+                        textStyle = MaterialTheme.typography.button.copy(
+                            if(state.isAttendeeEmailValid == true)
+                                MaterialTheme.colors.onSurface
+                            else
+                                MaterialTheme.colors.onSurface.copy(alpha = 0.3f)
+                        )
+                    ) {
+                        if(state.isAttendeeEmailValid == true) {
```
philipplackner, line 140: The comparison seems to not be necessary here (or is `isAttendeeEmailValid` nullable and if so why is it?)

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventViewModel.kt
```diff
@@ -171,7 +188,18 @@ class EventViewModel @Inject constructor(
...
+            }
+            is ValidateAttendeeEmail -> {
+                _state.update { _state ->
+                    _state.copy(
+                        isAttendeeEmailValid =
+                            if (uiEvent.email.isBlank())
+                                    null
+                                else
+                                    validateEmail.validate(uiEvent.email)
                     )
```
philipplackner, line 203: Ok seems like it is null, why not make it `false` by default, since an empty string is not a valid email?

realityexpander: Because I want to start the user out with a "clean, error-free" dialog state. And if its empty, there is no error yet. Only when the user starts typing do I indicate the validity of the text as an email address.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventEditors.kt
realityexpander: OK, i will check this.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventEditors.kt
realityexpander: I was able to make it work from the call site, but the MaterialTheme can only be accessed from the @Composable.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventEditors.kt
realityexpander: The reason it is necessary is because its null when the edit text field is blank. Once the user starts typing into the text field, it will be set to true or false and thats when I want to show the icon. 

I dont want to present the user with an error state before they even start typing. 

Does this make sense?

philipplackner: Ah, you do live validation. I thought this is just about whether the checkmark is displayed or not in the email field. In my app I only display an actual error message if the user tries to click on add when the email is invalid. But if you do it differently, it's fine

realityexpander: OK sounds good.

PR #20, opened 2022-11-17
0018 add photo chooser to event screen

realityexpander:
* Conform Cancel button behavior to match "tap-outside" and "cancel" button for Add Attendee dialog
* Add ability to add photos
* UI Lists both URL and URI photos
* Limits number of photos to 10
* Add sample photo URLs
* Use `PickVisualMediaRequest` to select photos

## Question 1

Is this a good & proper use of `derivedStateOf`:

https://github.com/realityexpander/Tasky/blob/8123fd3af6be2ba218b7274b2454615d025a7952/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventScreen.kt#L440-L474

### DIFF
--- a/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventEditors.kt
+++ b/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventEditors.kt
@@ -120,11 +120,6 @@ fun EventPropertyEditors(
         }
         is EditMode.ChooseAddPhoto -> {
             // Reference: https://github.com/philipplackner/NewPhotoPickerAndroid13/blob/master/app/src/main/java/com/plcoding/newphotopickerandroid13/MainActivity.kt
-
-            var selectedImageUri by remember {
-                mutableStateOf<Uri?>(null)
-            }
-
             singlePhotoPickerLauncher.launch(
                 PickVisualMediaRequest(ActivityResultContracts.PickVisualMedia.ImageOnly)
             )
--- a/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventEditors.kt
+++ b/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventEditors.kt
@@ -139,7 +139,6 @@ fun EventPropertyEditors(
                 buttons = {
                     positiveButton(
                         text = stringResource(R.string.ok),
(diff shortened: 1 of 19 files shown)


app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventEditors.kt
```diff
@@ -129,19 +139,22 @@ fun EventPropertyEditors(
                 buttons = {
                     positiveButton(
                         text = stringResource(R.string.ok),
-                        disableDismiss = true,
+                        disableDismiss = true,  // don't close the dialog automatically when OK button is tapped
```
philipplackner, line 142: Why not?

philipplackner (approved):
`derivedStateOf` seems overkill there. All the code you use to combine the list of photos could as well just be put in the ViewModel 🤔 
`derivedStateOf` helps if you have a state that is depending on another frequently changing state (such as `firstVisibleItemIndex` of a LazyColumn), but in your example the photos list doesn't change very often. I also don't understand why you need to keep track of 2 photo lists (`photos` and `photosToUpload`). You could just have one list and before uploading, you only send the local photos to the repo for the upload

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventEditors.kt
realityexpander: Because if there is an Error, i want the dialog to stay open. Without this flag, the OK button automatically closes the dialog, and I want it to stay open until the attendee is verified as existing and added.

realityexpander:
> `derivedStateOf` seems overkill there. All the code you use to combine the list of photos could as well just be put in the ViewModel 🤔 `derivedStateOf` helps if you have a state that is depending on another frequently changing state (such as `firstVisibleItemIndex` of a LazyColumn), but in your example the photos list doesn't change very often. I also don't understand why you need to keep track of 2 photo lists (`photos` and `photosToUpload`). You could just have one list and before uploading, you only send the local photos to the repo for the upload

OK, i will refactor to have a single list of photos. That makes more sense.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventEditors.kt
realityexpander: I will remove this at it seems to have no effect.

PR #21, opened 2022-11-18
0019 View Event Photo

realityexpander:
* Add View Photo dialog
* Add Remove photo button
* Add associated events
* Change var names for clarity

## No questions

### DIFF
--- a/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/components/PhotoModal.kt
+++ b/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/components/PhotoModal.kt
@@ -56,11 +56,11 @@ fun PhotoModal(
                     .padding(DP.small)
             ) {
 
-                // Back caret icon
+                // Close button
                 Icon(
                     Icons.Outlined.Close,
                     tint = MaterialTheme.colors.onSurface,
-                    contentDescription = "Back",
+                    contentDescription = stringResource(R.string.back_description),
                     modifier = Modifier
                         .size(32.dp)
                         .align(Alignment.CenterVertically)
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/components/PhotoModal.kt
```diff
@@ -0,0 +1,181 @@
...
+                    Icons.Outlined.Delete,
+                    tint = if (isRemoveEnabled) MaterialTheme.colors.onSurface else Color.Transparent,
+                    contentDescription = stringResource(R.string.remove_description),
+                    modifier = Modifier
+                        .size(32.dp)
+                        .align(Alignment.CenterVertically)
+                        .clickable(isRemoveEnabled) {
+                            isConfirmRemoveDialogVisible = true
+                        }
+                )
```
philipplackner, line 90: If you use clickable icons, I'd prefer to use `IconButton` instead which comes with all the default ripples and padding all icon buttons have on Android according to Material Design. It's something users are used to

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/components/PhotoModal.kt
```diff
@@ -0,0 +1,181 @@
...
+                        model = photo.url,
+                        contentDescription = stringResource(id = R.string.event_description_photo),
+                        modifier = Modifier
+                            .fillMaxSize()
+                            .weight(1f),
+                        contentScale = ContentScale.Fit
+                    )
+                }
+            }
+        }
```
philipplackner, line 118: Both these `AsyncImages` seem to be exactly the same except for the model. So here it would make more sense to instead use the `when` just for the model to get rid of needing 2 composables

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/components/PhotoModal.kt
realityexpander: ok i didnt know about that one. Will swap out.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/components/PhotoModal.kt
realityexpander: OK will swap out for when in the model.

PR #23, opened 2022-11-19
0021 delete event

realityexpander:
* [x] Refactor to have `AsyncImage` use single model
* [x] Change `Icon`s -> `IconButton`s
* [x] Improve dummy default data
* [x] Refactor `isUserEventCreator` "editable" items
  * [x] Only `RemindAt` can be edited by any attendee
* [x] Add "confirm delete `AgendaItem`"  dialog 
* [X] Add `Join` and `Leave` action placeholders to `Event` Screen
* [x] Added resource strings for `Event` screen dialogs

## Question 1
Can you toggle any `AgendaItem` or only `Tasks`?

## Question 2
Is it necessary to do a search in UI items for the `id` before deleting?
Or is it OK to simply delete the `AgendaItem` using the `id` that was passed in?

https://github.com/realityexpander/Tasky/blob/f7fe15dff26e5bfa31e77a78e2f9c5ab924156c5/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt#L212-L235

## Question 3
If a person `Leaves` an `Event`, does the event show up "crossed out" (like the `Task`-completed)? 
Or is there a different appearance in the `Agenda` screen list for events that the user is not going to?

## Question 4
It seems like Android does not let you save a `Uri` to a database and retrieve it later using `Uri.parse()` in such a way that you can get access to the original image. After parsing, most of the fields are empty and the access to the image fails. 

1 - Is this correct? Im not finding much information about how to handle this other than copying the contents of the Uri image to local cache and using saving the file to that cached image.
2 - Is the only way to make this work is to save the Uri image to local app cache as a new (duplicated) file?

https://github.com/realityexpander/Tasky/blob/7a3ba6e0faf00cf869ee98ce5c561990b1039e4c/app/src/main/java/com/realityexpander/tasky/agenda_feature/data/common/serializers/UriSerializer.kt#L14-L27


app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaState.kt
```diff
@@ -27,4 +27,6 @@ data class AgendaState(
 
     val agendaItems: Flow<List<AgendaItem>> = flow { emptyList<AgendaItem>() },
```
philipplackner, line 28: I find having a flow inside a state quite unintuitive, since it doesn't really feel like state (you can't just check the value of it without observing it). In this case, I'd probably prefer using a `combine()` block to combine your screen state with the flow you get from your repo and just have a normal list in the state class

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
```diff
@@ -134,8 +134,10 @@ class AgendaViewModel @Inject constructor(
                 }
             }
             is ToggleTaskCompleted -> {
-                _agendaState.value.agendaItems.mapLatest {
-                    it.find { item -> item.id == event.agendaItemId }?.let { task ->
+                _agendaState.value.agendaItems.collectLatest {
+                    it.find {
+                        item -> item.id == event.agendaItemId
+                    }?.let { task ->
```
philipplackner, line 140: Why do you use a collect block in here? If this gets called, the collector will block this function until it stopped collecting, I don't think this is what you want 🤔 
I'd either pass the Task that was clicked on from the UI or simply use `_agendaState.value.agendaItems.find {}`

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
```diff
@@ -193,6 +195,14 @@ class AgendaViewModel @Inject constructor(
                 }
                 sendEvent(ShowProgressIndicator(false))
             }
+            is ClearErrorMessage -> {
+                _agendaState.update {
+                    it.copy(
+                        errorMessage = null
+                    )
+                }
+                sendEvent(ShowProgressIndicator(false))
```
philipplackner, line 204: Why is it an event to show a progress indicator? I'd consider this a normal state

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
```diff
@@ -202,6 +212,59 @@ class AgendaViewModel @Inject constructor(
             is OneTimeEvent.NavigateToEditEvent -> {
                 _oneTimeEvent.emit(OneTimeEvent.NavigateToEditEvent(event.eventId))
             }
+            is ShowConfirmDeleteAgendaItemDialog -> {
+                _agendaState.update {
+                    it.copy(
+                        confirmDeleteAgendaItem = event.agendaItem
+                    )
+                }
+                sendEvent(ClearErrorMessage)
```
philipplackner, line 221: Same here, wouldn't it make more sense to just update the state with let's say `errorMessage = null`?

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
```diff
@@ -202,6 +212,59 @@ class AgendaViewModel @Inject constructor(
...
+            }
+            is DeleteAgendaItem -> {
+                // Question: Is it necessary to do a search in UI items for the ID before deleting?
+                //   Or is it OK to simply delete the AgendaItem that was passed in?
+                _agendaState.value.agendaItems.collectLatest {
```
philipplackner, line 226: It's fine to just take the element that was passed in

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
```diff
@@ -261,18 +324,33 @@ class AgendaViewModel @Inject constructor(
...
                     description = "Discuss the yet another project",
                     host = "Jack",
                     isUserEventCreator = true,
                     isGoing = false,
-                    attendees = emptyList(),
+                    attendees = listOf(
+                        Attendee(
+                            id = authRepository.getAuthInfo()?.userId!!,
+                            fullName = authRepository.getAuthInfo()?.username ?: "",
+                            email = authRepository.getAuthInfo()?.email ?: "",
```
philipplackner, line 350: Will this read from preferences 3 times here?

philipplackner (approved):
I commented twice on something regarding your events in the VM, but now after reviewing I understand what you're doing. I'm still wondering what's the benefit of sending an event which the VM listens to to update the state instead of just calling a function which updates the state?

1. No, only Tasks.
2. Answered in comment
3. If a person leaves an event, they won't be able to see it anymore. The `/agenda` route responds with all events you either own or are attendee of. If you leave it, you're neither of that.
4. What do you need to save the Uri for? Editing photos is only supported in online mode. If you'd want that, you'd need to save the photo in the app's cache, yes. Even if you could just save the Uri, the user could delete the photo and make it invalid that way. I wouldn't consider saving this as a copy here as a problem, since phones are online quite frequently, and after syncing it can be deleted again.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
realityexpander: The preference is in RAM at this time, so the access is very fast. 

Only when writing it it is slow.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
realityexpander: OK, I will refactor to just use the `id` passed in.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
realityexpander: > I commented twice on something regarding your events in the VM, but now after reviewing I understand what you're doing. I'm still wondering what's the benefit of sending an event which the VM listens to to update the state instead of just calling a function which updates the state?

I just wanted a centralized place where all the events are processed and updated, and not a bunch of separate functions. I'm mimickying the MVI pattern a bit here. 

## Follow up question

 Is this a bad practice?

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaState.kt
realityexpander: OK, I will refactor to use a normal list in the state class.

realityexpander: Im not sure I follow you here.

I tried to use your recommendation of `.combine()` and was successful in making the `_agendaState` hold the flow result as a single list, but when I tried to change the `selectedDayIndex` and use the result as a flow, I was not able to update the `agendaDayFlow`. 

Am I understanding that I need to update the `AgendaItems` statically in the `.combine()` and not use a flow for the `selectedDay`? I have updated the `AgendaViewModel` code to reflect this change, but not sure if this approach is the best. 

## Follow up question
Is this is the correct solution you were referring to?

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
realityexpander: OK, i will refactor this.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
realityexpander: OK I will refactor.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
realityexpander: Since its not doing an state `update` here, and just clearing the error message, i centralized it to also get rid of any progress indicator too using an event.

realityexpander:
> 4\. What do you need to save the Uri for? Editing photos is only supported in online mode. If you'd want that, you'd need to save the photo in the app's cache, yes. Even if you could just save the Uri, the user could delete the photo and make it invalid that way. I wouldn't consider saving this as a copy here as a problem, since phones are online quite frequently, and after syncing it can be deleted again.

OK that is what I thought. I will attempt to implement the "offline adding" of photos.

## Follow up question

I would have thought that the URI would point to the file and if it had not moved or been deleted it would still be accessible. But looking online this does not seem to be the case for URI's that are saved as strings and then re-parsed back to URI's. 

Is this correct?

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaState.kt
philipplackner: Sorry for the confusion! The main issue I'm talking about here is that you will have no way of accessing the current agenda items without launching a new flow collector (even if it's just done via `.first()`, you'll have another DB query for that). 

What would make sense in my head is a setup like this. Might need a bit of adjustment with savedStateHandle and such, but the idea is that we just flatten the flow from the repo (so we get rid of the Flow<*> wrapper and just get the list emissions which we can use to update the agenda state):
```
private val _selectedDayIndex = MutableStateFlow<Int?>(null) // Can remain private
private val agendaItems = _selectedDayIndex
    .flatMapLatest { index ->
        agendaRepository.getAgendaForDayFlow(
            getDateForSelectedDayIndex(index)
        )
    }
    .stateIn(viewModelScope, SharingStarted.WhileSubscribed(5000), emptyList())

private val _agendaState = MutableStateFlow(
    AgendaState(
        errorMessage = errorMessage,
        selectedDayIndex = selectedDayIndex,
))
val agendaState = combine(
    _agendaState,
    agendaItems
) { state, items -> 
    // If performance is super important, you could also check here if the list changed and only then copy it
    state.copy(
        agendaItems = items
    )
}.stateIn(viewModelScope, SharingStarted.WhileSubscribed(5000), AgendaState())
```

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
philipplackner: If that object was already read from memory and just sits in a local variable, your approach is perfectly fine. If new DataStore reads are happening on every call, I'd simply save it in one 👍 

2. question:
Hm, not sure I'd call it a bad practice, it just doesn't feel intuitive to me to go that extra mile if you can just update the state right away using a normal function that expresses the same as your event name. Then you don't need any extra collectors etc.

philipplackner:
I think it depends on the type of Uri.. In this app you're dealing with content Uris, for which this is true I think. On the other hand there are file Uris for example for which I don't think this is true, since they rather refer to a file path (in your app's storage)

realityexpander:
> I think it depends on the type of Uri.. In this app you're dealing with content Uris, for which this is true I think. On the other hand there are file Uris for example for which I don't think this is true, since they rather refer to a file path (in your app's storage)

OK thats the conclusion I came to as well and from research online.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaState.kt
realityexpander: OK, i have refactored this.

PR #24, opened 2022-11-20
0022 Upload Created Event and Photos

realityexpander:
* Use serializer with encoded string
* Add error message handling
* get data using bytes to automatically recompress an image
* add progress indicators
* fixed list of strings <-> object issue, now allow empty strings convert to empty List.
* make pretty printing easier to read
* naming improvements

### DIFF
--- a/app/src/androidTest/java/com/realityexpander/tasky/EventDaoTest.kt
+++ b/app/src/androidTest/java/com/realityexpander/tasky/EventDaoTest.kt
@@ -67,7 +67,7 @@ open class EventDaoTest {
             photos = listOf(
                 PhotoEntity("1", "https://www.google.com")
             ),
-            deletedPhotoKeys = emptyList(),
+            deletedPhotoIds = emptyList(),
             isDeleted = false,
         )
 
@@ -112,7 +112,7 @@ open class EventDaoTest {
             photos = listOf(
                 PhotoEntity("1", "https://www.google.com")
             ),
-            deletedPhotoKeys = listOf(),
+            deletedPhotoIds = listOf(),
             isDeleted = false,
         )
 
@@ -160,7 +160,7 @@ open class EventDaoTest {
             photos = listOf(
                 PhotoEntity("1", "https://www.google.com")
             ),
-            deletedPhotoKeys = listOf(),
+            deletedPhotoIds = listOf(),
             isDeleted = false,
         )
 
@@ -206,7 +206,7 @@ open class EventDaoTest {
             photos = listOf(
                 PhotoEntity("1", "https://www.google.com")
             ),
-            deletedPhotoKeys = listOf(),
+            deletedPhotoIds = listOf(),
             isDeleted = false,
         )
         val event2 = EventEntity(
@@ -233,7 +233,7 @@ open class EventDaoTest {
             photos = listOf(
                 PhotoEntity("1", "https://www.google.com")
             ),
-            deletedPhotoKeys = listOf(),
+            deletedPhotoIds = listOf(),
             isDeleted = false,
         )
 
@@ -280,7 +280,7 @@ open class EventDaoTest {
             photos = listOf(
                 PhotoEntity("1", "https://www.google.com")
             ),
-            deletedPhotoKeys = listOf(),
+            deletedPhotoIds = listOf(),
             isDeleted = false,
         )
         val event2 = EventEntity(
@@ -307,7 +307,7 @@ open class EventDaoTest {
             photos = listOf(
                 PhotoEntity("1", "https://www.google.com")
             ),
-            deletedPhotoKeys = listOf(),
+            deletedPhotoIds = listOf(),
             isDeleted = false,
         )
 
@@ -360,7 +360,7 @@ open class EventDaoTest {
             photos = listOf(
                 PhotoEntity("1", "https://www.google.com")
             ),
-            deletedPhotoKeys = listOf(),
+            deletedPhotoIds = listOf(),
             isDeleted = false,
         )
         val event2 = EventEntity(
@@ -387,7 +387,7 @@ open class EventDaoTest {
             photos = listOf(
                 PhotoEntity("1", "https://www.google.com")
             ),
-            deletedPhotoKeys = listOf(),
+            deletedPhotoIds = listOf(),
             isDeleted = false,
         )
 
@@ -439,7 +439,7 @@ open class EventDaoTest {
             photos = listOf(
                 PhotoEntity("1", "https://www.google.com")
             ),
-            deletedPhotoKeys = listOf(),
+            deletedPhotoIds = listOf(),
             isDeleted = false,
         )
         val event2 = EventEntity(
@@ -466,7 +466,7 @@ open class EventDaoTest {
             photos = listOf(
                 PhotoEntity("1", "https://www.google.com")
             ),
-            deletedPhotoKeys = listOf(),
+            deletedPhotoIds = listOf(),
             isDeleted = false,
         )
 
@@ -525,7 +525,7 @@ open class EventDaoTest {
             photos = listOf(
                 PhotoEntity("1", "https://www.google.com")
             ),
-            deletedPhotoKeys = listOf(),
+            deletedPhotoIds = listOf(),
             isDeleted = false,
         )
         val event2 = EventEntity( // This event is not for today
@@ -552,7 +552,7 @@ open class EventDaoTest {
             photos = listOf(
                 PhotoEntity("1", "https://www.google.com")
             ),
-            deletedPhotoKeys = listOf(),
+            deletedPhotoIds = listOf(),
             isDeleted = false,
         )
         val event3 = EventEntity(  // This event is not for today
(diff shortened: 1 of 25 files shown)


app/src/androidTest/java/com/realityexpander/tasky/EventDaoTest.kt
```diff
@@ -67,7 +67,7 @@ open class EventDaoTest {
             photos = listOf(
                 PhotoEntity("1", "https://www.google.com")
             ),
-            deletedPhotoKeys = emptyList(),
+            deletedPhotoIds = emptyList(),
```
philipplackner, line 70: In practice, it totally makes sense to put such renamings and refactorings that affect a lot of classes into a separate PR (which doesn't need review for this app), since these things quickly consume half of a PR and you gotta spot the important changes in between

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/common/convertersDTOEntityDomain/EventConverters.kt
```diff
@@ -124,8 +132,7 @@ fun AgendaItem.Event.toEventDTOResponse(): EventDTO.Response {
         isGoing = isGoing ?: false,
         photos = photos
             .map { it.toDTO() }
-            .filterIsInstance<PhotoDTO.Remote>()        // ensure only respond with Remote photos
-        ,
+            .filterIsInstance<PhotoDTO.Remote>(),        // ensure only respond with Remote photos
```
philipplackner, line 135: In which case would a `PhotoDTO` contain local photos?

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/eventRepository/remote/eventApi/DTOs/EventDTO.kt
```diff
@@ -67,7 +68,7 @@ abstract class EventDTO {
 
         val host: UserId,
         val isUserEventCreator: Boolean,
-        val isGoing: Boolean? = null,
+        val isGoing: Boolean? = true,
```
philipplackner, line 71: Is there any case where this would be `null`? If not, you can as well make it non-nullable

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/eventRepository/remote/eventApi/eventApiImpls/EventApiImpl.kt
```diff
@@ -1,51 +1,72 @@
...
-//                }.also {
-//                }
+                photos = event.photos.mapIndexed { index, photo -> // todo - add photo handling
+
+                    val bytes = getBytesFromUri(context, photo.uri)?.also { bytes ->
+                        val size = bytes.size
+                        if (size > MAX_IMAGE_SIZE) {
+                            throw NetworkErrorException("Photo too large")
+                        }
+                    } ?: throw Exception("Photo not found")
```
philipplackner, line 55: If the user tries to upload 10 photos and just one is too large, this error won't tell them a lot about what they should do (since they don't know which one is too large). I think one of these make sense:
1. Check the photo size when the user tries to add it (maybe you already do that, then ignore this comment). Downside: You'll need to read the bytes of it before you know you'll need them for the upload (however, the size might also be contained in the metadata, not sure right now)
2. You only upload those photos that are actually small enough and let the user know something like `Some photos were too large to be uploaded and were ignored`

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/eventRepository/remote/eventApi/eventApiImpls/EventApiImpl.kt
```diff
@@ -66,43 +87,94 @@ class EventApiImpl @Inject constructor(
...
+                it?.readBytes()
+            }
+
+        val imageSize = context.contentResolver.openFileDescriptor(uri, "r")
+            .use {
+                it?.statSize
+            }
+
+        imageSize?.also { size ->
+            if (size > 1000000) {
```
philipplackner, line 167: Should use your `MAX_IMAGE_SIZE` constant I guess

app/src/main/java/com/realityexpander/tasky/core/util/getRealPathFromURI.kt
```diff
@@ -0,0 +1,152 @@
...
+    return "com.android.providers.media.documents" == uri.authority
+}
+
+/**
+ * @param uri The Uri to check.
+ * @return Whether the Uri authority is Google Photos.
+ */
+fun isGooglePhotosUri(uri: Uri): Boolean {
+    return "com.google.android.apps.photos.content" == uri.authority
+}
```
philipplackner, line 152: I wonder where you need all these functions 🤔 
Also, this should go in data instead of presentation

app/src/androidTest/java/com/realityexpander/tasky/EventDaoTest.kt
realityexpander: OK, I will move this kind of activity to a separate PR

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/eventRepository/remote/eventApi/DTOs/EventDTO.kt
realityexpander: OK will use non-nullable

app/src/main/java/com/realityexpander/tasky/core/util/getRealPathFromURI.kt
realityexpander: I will remove unneeded code. This was some code I was working with to get better handle on file uploading

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/common/convertersDTOEntityDomain/EventConverters.kt
realityexpander: It should never contain them, this is just a sanity check, and also conveying intent about what is going on. 

Is this a bad practice?

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/eventRepository/remote/eventApi/eventApiImpls/EventApiImpl.kt
realityexpander: OK, I will refactor this for better error handling.

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/eventRepository/remote/eventApi/eventApiImpls/EventApiImpl.kt
realityexpander: OK, will use the constant.

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/common/convertersDTOEntityDomain/EventConverters.kt
philipplackner: No it of course makes sense to make sure only local photos can be uploaded, but I'm wondering why you don't just model your `PhotoDTO` like this to avoid that local photos can be passed at all:
```
data class PhotoDTO(
    val photos: List<PhotoDTO.Remote>
)
```

philipplackner: Yes showing intent is great, I just don't know if I'd do it by implementing an unused class here (`PhotoDTO.Local`). What do you think about something like this?
```
data class RemotePhotoDTO(
    val url: String, 
    val key: String
)

data class EventDTO(
    val photos: List<RemotePhotoDTO>
    ...
)
```

realityexpander: OK, maybe I was mistaken about what the PhotoDTO is used for. 

`PhotoDTO.Local` contains the URIs that are decoded and sent to the server. 

When a api response come back, the `PhotoDTO.Remote` is populated with links to server images. You cannot convert a `PhotoDTO.Local` to a domain or entity object.

philipplackner: Yeah you can do it like that :)
I was just curious, since your data layer could also perfectly access a model from domain (like `Photo.Local` here), but I see the point of following consistency here.

realityexpander: OK cool!

PR #26, opened 2022-11-24
0024 Improve Image Handling for images too large to upload

realityexpander:
* Refactor to use non-null type for domain, DTO & Entity for Event
* Reset database scheme
* move uri handler to data package
* Handle the "too large to upload" message when at the image picker UI, not when attempting to upload it.

## Question 1
Should I display an Alert when an image is too large to upload? Or is an error message or snackbar a better place for this kind of error message?

Relevant code:
https://github.com/realityexpander/Tasky/blob/abba42677130de4d83e01a7f84ef6d8e3c297291/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventScreen.kt#L141-L165

## Question 2

I mis-named this branch when I was 10% into the work, and it seems like re-naming a branch is not a straightforward mechanism. Looks like you have to create and delete a branch. 

Is there a straightforward way to do this from the AS UI?

### DIFF
--- a/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventScreen.kt
+++ b/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventScreen.kt
@@ -44,9 +44,12 @@ import com.realityexpander.tasky.agenda_feature.presentation.event_screen.compon
 import com.realityexpander.tasky.agenda_feature.presentation.event_screen.components.SmallHeightHorizontalDivider
 import com.realityexpander.tasky.agenda_feature.util.toLongMonthDayYear
 import com.realityexpander.tasky.auth_feature.domain.AuthInfo
+import com.realityexpander.tasky.core.data.isImageSizeTooLargeToUpload
 import com.realityexpander.tasky.core.presentation.common.modifiers.*
+import com.realityexpander.tasky.core.presentation.common.util.UiText
 import com.realityexpander.tasky.core.presentation.theme.TaskyLightGreen
 import com.realityexpander.tasky.core.presentation.theme.TaskyTheme
+import com.realityexpander.tasky.core.util.UPLOAD_IMAGE_MAX_SIZE
 import com.realityexpander.tasky.core.util.UuidStr
 import kotlinx.coroutines.launch
 import java.time.ZonedDateTime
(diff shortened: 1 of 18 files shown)


app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventScreen.kt
```diff
@@ -139,8 +142,15 @@ fun AddEventScreenContent(
     val singlePhotoPickerLauncher =
         rememberLauncherForActivityResult(
             contract = ActivityResultContracts.PickVisualMedia(),
-            onResult = { uri ->
-                uri?.let {
+            onResult = { uriNullable ->
+
+                uriNullable?.let { uri ->
+
+                    if(uri.isImageSizeTooLargeToUpload(context, UPLOAD_IMAGE_MAX_SIZE)) {
```
philipplackner, line 149: Here, I'd actually not make this an extension function of `Uri`, since I'd then expect this to be generic behavior that applies to all Uris, but it only applies to event photo Uris. This logic is rather specific to AgendaItems/Events, so maybe a function in a companion object in one of these classes could make more sense?
```
Event.isImageSizeTooLargeToUpload(uri, context, UPLOAD_IMAGE_MAX_SIZE)
```

Or you create something like a `UriByteConverter`:
```
// In domain
interface UriByteConverter {
    fun convertToBytes(uri: Uri, maxSize: Long): ByteArray
}

// In data
class UriByteConverter(private val context: Context): UriByteConverter {
    override fun convertToBytes(uri: Uri, maxSize: Long): ByteArray {
        ...
    }
}
```

app/src/main/java/com/realityexpander/tasky/core/data/getBytesFromUri.kt
```diff
@@ -0,0 +1,44 @@
...
+            val stream = ByteArrayOutputStream()
+            bitmap.compress(Bitmap.CompressFormat.JPEG, 50, stream)
+            bytes = stream.toByteArray()
+        }
+    }
+
+    return bytes
+}
+
+fun Uri.isImageSizeTooLargeToUpload(context: Context, maxUploadSize: Int = UPLOAD_IMAGE_MAX_SIZE): Boolean {
```
philipplackner, line 38: Okay, now when seeing the function, it seems fine, just that I would rename it to something more generic since it might be used for more than just checking photo size before upload. Maybe just `Uri.getSize(): Long`?
Then you're more flexible where you use this function for what

philipplackner (commented):
1. I think an alert dialog is justified in this case, since it's quite an important info for the user why they can't update the event (or why certain photos weren't uploaded)
2. I don't think you can easily do that after pushing your stuff. If it's a local branch, you can easily do it in AS. Happens and is not the end of the world, I don't even look at your branch names, just at the PR description.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventScreen.kt
realityexpander: OK will refactor this for correct package

app/src/main/java/com/realityexpander/tasky/core/data/getBytesFromUri.kt
realityexpander: I have refactored these for clarity and put into appropriate packages.

PR #2, opened 2022-10-27
0002 login register UI

realityexpander:
- Repo, API and Dao are all still using fakes
- Improves MVI-style event handling, using Single Source of Truth in the SavedStateHandle
- Completed Login Screen & ViewModel Logic
- Completed Registration Screen & ViewModel Logic
- Wrote a few UI Test and ViewModel tests (i messaged you about one problem with ViewModel test on Login)
- Improved your UiText class for more flexibility and usability
- Completed basic splash screen with Tasky Logo
- Added ValidateUsername usecase
- Improved components for EmailField, Password field by implementing abstracted TextEntryField
- Bumped various library versions and using Kotlin 1.6.21 now.

## Question 1:
Is this too big of a PR? Should I be doing smaller PR's? I feel all these function together as a unit, so I thought it would be better to include as one PR.

## Question 2:
When I am updating the MutableStateFlow (like `_loginState` in `LoginViewModel`) should I use the derived `loginState` as the source of data to `.copy()` & update, or use the Mutable `_loginState` as the source? Currently i am using the Mutable `_loginState` as the source of truth.

## Question 3: (from the whatsapp)
Are there any recommended resources for making instrumented UI tests for compose where we can send events and be updated in the UI and tests can be done on ui elements, similar to how espresso did for XML?

From the documentation, it seems possible to send events to my `ViewModel` but my tests only respond to the first emission from the `MutableStateFlow`. Subsequent events sent from inside my `@Test` are not responded to. You can check out my test in for the `LoginViewModel` in the regular `test` folder.

PR #27, opened 2022-11-24
0025-add-agenda-date-picker-2

realityexpander:
* Make agenda date picker operate properly
* add strings for yesterday, today and tomorrow
* Add date picker dialog for current day

## Question 1
Is this a good way to handle combining two flows?

https://github.com/realityexpander/Tasky/blob/39dabd036e19bc9b7250332e66e390f6822c4761/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt#L49-L57

### DIFF
--- a/app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/eventRepository/remote/eventApi/eventApiImpls/EventApiImpl.kt
+++ b/app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/eventRepository/remote/eventApi/eventApiImpls/EventApiImpl.kt
@@ -152,31 +152,4 @@ class EventApiImpl @Inject constructor(
                     ?: "jpg")
     }
 
-//    // Will automatically recompress to lower quality if the image is too large
-//    private fun getBytesFromUri(context: Context, uri: Uri): ByteArray?  {
-//        var bytes = context.contentResolver
-//            .openInputStream(uri)
-//            .use {
-//                it?.readBytes()
-//            }
-//
-//        val imageSize = context.contentResolver.openFileDescriptor(uri, "r")
-//            .use {
-//                it?.statSize
-//            }
-//
-//        imageSize?.also { size ->
-//            if (size > MAX_IMAGE_SIZE) {
-//                //throw NetworkErrorException("Photo too large")
-//
-//                // recompress the photo
-//                val bitmap = BitmapFactory.decodeByteArray(bytes, 0, bytes?.size ?: 0)
-//                val stream = ByteArrayOutputStream()
-//                bitmap.compress(Bitmap.CompressFormat.JPEG, 50, stream)
-//                bytes = stream.toByteArray()
-//            }
-//        }
-//
-//        return bytes
-//    }
 }
--- a/app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/eventRepository/remote/eventApi/eventApiImpls/EventApiImpl.kt
+++ b/app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/eventRepository/remote/eventApi/eventApiImpls/EventApiImpl.kt
@@ -5,6 +5,7 @@ import android.content.Context
 import android.net.Uri
 import com.realityexpander.tasky.agenda_feature.common.util.EventId
 import com.realityexpander.tasky.agenda_feature.data.repositories.eventRepository.remote.eventApi.DTOs.EventDTO
+import com.realityexpander.tasky.agenda_feature.data.repositories.eventRepository.remote.eventApi.DTOs.PhotoDTO
 import com.realityexpander.tasky.agenda_feature.data.repositories.eventRepository.remote.eventApi.IEventApi
 import com.realityexpander.tasky.core.data.getBytesFromUri
 import com.realityexpander.tasky.core.data.remote.TaskyApi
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/eventRepository/remote/eventApi/eventApiImpls/EventApiImpl.kt
```diff
@@ -138,6 +104,24 @@ class EventApiImpl @Inject constructor(
...
     }
 
+    ////////////////// HELPER FUNCTIONS //////////////////
+
+    private fun getMultipartBodyPartsForLocalPhotos(
+        photos: List<PhotoDTO.Local>
+    ): List<MultipartBody.Part> {
+        return photos.mapIndexed { index, photo ->
+            val bytes = photo.uri.getBytesFromUri(context, UPLOAD_IMAGE_MAX_SIZE)
+                ?: throw Exception("Photo not found")
```
philipplackner, line 114: Do you catch this somewhere? What if the user adds a photo, then deletes it from their device, goes back to the app and hits save?

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaScreen.kt
```diff
@@ -614,6 +625,39 @@ fun AgendaScreenContent(
...
+                positiveButton(text = stringResource(R.string.ok)) {
+                    dateDialogState.hide()
+                    onAction(SetCurrentDate(pickedDate))
+                }
+                negativeButton(text = stringResource(R.string.cancel)) {
+                    dateDialogState.hide()
+                    onAction(CancelChooseCurrentDateDialog)
+                }
+            }
+        ) {
```
philipplackner, line 651: This looks like a side effect. `dateDialogState.show()` is not a composable function, but executed in a composable context. You can safely put the dialog composable anywhere at the root of your `AgendaScreen` composable and show it like this:
```
LaunchedEffect(state.chooseCurrentDateDialog) {
    state.chooseCurrentDateDialog?.let { dateDialogState.show() }
}
```

philipplackner (commented):
Answer to your question:
Yeah, combining looks fine. It looks a bit weird to use `flatMapLatest` without mapping anything, but I think it's the only way, since there doesn't seem to be a `flatten()` function for flows (as there is for lists).
I found this: https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/flatten-merge.html
But I can't find that function in my project, not sure if it's only available for a higher Kotlin version or so 🤔 

Just one thing that isn't clear to me in that block is why `getDateForSelectedDateIndex()` is needed. If `agendaRepository.getAgendaForDayFlow()` takes in a date, why can't you just pass `_currentDate`? Not sure if `currentDate` reflects the current or selected date for you. If it reflects the current one, it doesn't seem to be needed as a state, since you can always get that using `LocalDate.now()`

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/eventRepository/remote/eventApi/eventApiImpls/EventApiImpl.kt
realityexpander: Good point, I will put some error handling in for this.

realityexpander:
> Just one thing that isn't clear to me in that block is why `getDateForSelectedDateIndex()` is needed. If `agendaRepository.getAgendaForDayFlow()` takes in a date, why can't you just pass `_currentDate`? Not sure if `currentDate` reflects the current or selected date for you. If it reflects the current one, it doesn't seem to be needed as a state, since you can always get that using `LocalDate.now()`

Because I have a "anchor" date that is the start of the week for the agendaScreen, then I have a selector that sets the "index day". This makes handling the UI very straightforward.

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaScreen.kt
realityexpander: OK, I will refactor this.

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/eventRepository/remote/eventApi/eventApiImpls/EventApiImpl.kt
realityexpander: ## FOLLOW UP QUESTION:

I tested this case of deleted photo, and it currently gives a system level error that is coming from the `Uri` handling functions. 

I see 2 approaches:

1. When the user selects the photos, copy it to the app cache and then upload the file data from there and delete when uploaded.
2. When the user presses `save`, check the Uri's for validity and if all are not valid, show error dialog and then remove the photo(s) that are no longer valid.

I will implement #2 because its easier for now, but wondering if #1 is better long term.

philipplackner: I think either is fine and each has its up- and downside. The first won't lead to errors, but increases app storage and complexity (also things like what if there's no space left, which might confuse a user since they'll think the photo is already saved on their device) while the latter is easier, but comes with an error. In the end it's really just an edge case and even if this was an app with lots of users maybe 1% of them would ever face this.

realityexpander: OK, i went for Option 1, as its easier to implement, an edge case, and small impact. 

But at least the error is not cryptic, and the edge case is covered.

For more serious type of online data gathering, i would cache everything, just in case.

PR #30, opened 2022-11-26
0028 error handling for deleted local photos

realityexpander:
* [X] check for file existence before save instead of at upload time
* [x] Add German and Spanish translations (used a free tool)
  * note: most of this PR is these translations 
* [X] Add UI to handle telling use about missing photos
* [X] remove old comments


philipplackner (commented):
Good job 👍

PR #31, opened 2022-11-26
0029 delete attendee, leave/join/delete functionality, logout fix

realityexpander:
* [x] Add ability to remove Attendee from event (as current logged-in user)
* [x] Add ability to delete Event
* [x] Add simple on-line synchronization with server
* [x] Small API changes to use `@Query` instead of `@Path`
* [x] Add new resource strings
* [X] Improve AlertDialog to close automatically
* [X] Improve return error handling 
* [X] Fix logout crash
* [X] Refactor to implement refined understanding of `attendee` leave/join flow
* [X] clear database upon logout
* [X] correct handling of `isGoing` flag
* [X] Toggle `Attendee->isGoing` on the Event screen.
   * and only send the `updateEvent` to the server when the user `save`s the event.
* [X] Removed extraneous `isGoing` for `Event` (not `Attendee`) 

## Question 1
I tried using your `.flattenMerge()` but was not getting all events, so i switched back to `.flatmapLatest{ it }` and it works fine. Not sure whats going on here...

https://github.com/realityexpander/Tasky/blob/cfa78ba3ae5fb874c6c5a7fed7fc36cc3d00a296/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt#L50-L58

### DIFF
--- a/app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
+++ b/app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
@@ -7,7 +7,6 @@ import com.realityexpander.tasky.agenda_feature.data.common.convertersDTOEntityD
 import com.realityexpander.tasky.agenda_feature.data.repositories.agendaRepository.remote.IAgendaApi
 import com.realityexpander.tasky.agenda_feature.data.repositories.attendeeRepository.IAttendeeRepository
 import com.realityexpander.tasky.agenda_feature.domain.*
-import com.realityexpander.tasky.auth_feature.domain.AuthInfo
 import com.realityexpander.tasky.core.presentation.common.util.UiText
 import com.realityexpander.tasky.core.util.Email
 import kotlinx.coroutines.CancellationException
@@ -86,8 +85,8 @@ class AgendaRepositoryImpl @Inject constructor(
         return eventRepository.getEvent(eventId)
     }
 
-    override suspend fun updateEvent(event: AgendaItem.Event, authInfo: AuthInfo): ResultUiText<AgendaItem.Event> {
-        return eventRepository.updateEvent(event, authInfo.userId ?: throw java.lang.IllegalStateException("User id is null"))
+    override suspend fun updateEvent(event: AgendaItem.Event): ResultUiText<AgendaItem.Event> {
+        return eventRepository.updateEvent(event)
     }
 
     override suspend fun deleteEventId(eventId: EventId): ResultUiText<Void> {
--- a/app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
+++ b/app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
@@ -3,12 +3,17 @@ package com.realityexpander.tasky.agenda_feature.data.repositories.agendaReposit
 import com.realityexpander.tasky.R
 import com.realityexpander.tasky.agenda_feature.common.util.EventId
 import com.realityexpander.tasky.agenda_feature.data.common.convertersDTOEntityDomain.toDTO
+import com.realityexpander.tasky.agenda_feature.data.common.convertersDTOEntityDomain.toDomain
 import com.realityexpander.tasky.agenda_feature.data.repositories.agendaRepository.remote.IAgendaApi
 import com.realityexpander.tasky.agenda_feature.data.repositories.attendeeRepository.IAttendeeRepository
 import com.realityexpander.tasky.agenda_feature.domain.*
 import com.realityexpander.tasky.core.presentation.common.util.UiText
 import com.realityexpander.tasky.core.util.Email
+import kotlinx.coroutines.CancellationException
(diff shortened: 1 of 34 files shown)


app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
```diff
@@ -28,14 +33,31 @@ class AgendaRepositoryImpl @Inject constructor(
...
+                // don't send error to user, just log it (silent fail is ok here)
+            }
+        }
+
         val events = eventRepository.getEventsForDayFlow(dateTime)
-//        val tasks = taskRepository.getTasks(dateTime)                             // todo implement tasks repo
-//        val reminders = reminderRepository.getReminders(dateTime)                 // todo implement reminders repo
+    //        val tasks = taskRepository.getTasks(dateTime)                             // todo implement tasks repo
+    //        val reminders = reminderRepository.getReminders(dateTime)                 // todo implement reminders repo
         return events // + tasks + reminders
```
philipplackner, line 56: Instead of launching a new coroutine in a repo (which allows you 0 control from the outside about the execution of this), I'd make put this in one flow builder similar like this:
```
return flow {
    supervisorScope {
        launch { // your API call }

        emitAll(eventRepository.getEventsForDayFlow(dateTime))
    }
}
```

This way when canceling this, the API call also gets cancelled as you would expect

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/eventRepository/eventRepositoryImpls/EventRepositoryImpl.kt
```diff
@@ -36,6 +38,24 @@ class EventRepositoryImpl(
...
+
+    override suspend fun upsertEventLocally(event: AgendaItem.Event): ResultUiText<Void> {
+        return try {
+            eventDao.upsertEvent(event.toEntity())  // save to local DB ONLY
+
+            ResultUiText.Success()
+        } catch (e: CancellationException) {
+            throw e
+        } catch (e: Exception) {
+            ResultUiText.Error(UiText.Str(e.message ?: "upsertEvent error"))
```
philipplackner, line 52: What could go wrong here?

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventViewModel.kt
```diff
@@ -472,8 +482,39 @@ class EventViewModel @Inject constructor(
...
+                                } else {
+                                    it
+                                }
+                            },
+                            isGoing = true
+                        )
+
+                    )
+                }
+            }
```
philipplackner, line 517: Could make sense to create a function `setIsGoingForLocalUser(isGoing)` to avoid this duplication 😄

philipplackner (approved):

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/eventRepository/eventRepositoryImpls/EventRepositoryImpl.kt
realityexpander: Out of space on the device. Maybe some kind of configuration error? 

Is there a better way to handle this?

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/event_screen/EventViewModel.kt
realityexpander: ok, will refactor this.

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
realityexpander: ## FOLLOW UP QUESTION 1

> "This way when canceling this,..."

Not sure which `this` you are referring to. Do you mean the `supervisorscope` or the `flow`?

philipplackner: With `this` I mean the coroutine in which the flow is launched. When the flow is cancelled, the coroutine launched in `CoroutineScope(Dispatchers.IO)` will continue running since it's a different scope.

Yes, that's the implementation I meant

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/eventRepository/eventRepositoryImpls/EventRepositoryImpl.kt
philipplackner: I think I've never seen Room throw an exception in my life, but out of storage is a valid point, so please ignore this comment 👍

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
realityexpander: OK, its becoming clearer

PR #32, opened 2022-11-28
0030 handle deleted photos

realityexpander:
* Add ability to remove Photo (remote only)
* Add easier editing of titles and description
* Refresh from api

## QUESTION 1 (sorry if repeat)
Does `getAgenda` endpoint only get agenda items that are within that day? 
Can agenda items straddle more than one day? (my code is setup for that)

### DIFF
--- a/app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
+++ b/app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
@@ -43,9 +43,8 @@ class AgendaRepositoryImpl @Inject constructor(
 
                         // Get fresh data
                         val result = agendaApi.getAgenda(dateTime)
+//                        eventRepository.clearEventsForDayLocally(dateTime) // clear local data
                         result.events.forEach { event ->
-                            eventRepository.clearEventsForDayLocally(dateTime) // clear local data
-
                             // Insert fresh data into db
                             val result2 =
                                 eventRepository.upsertEventLocally(event.toDomain())
--- a/app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
+++ b/app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
@@ -10,9 +10,9 @@ import com.realityexpander.tasky.agenda_feature.domain.*
 import com.realityexpander.tasky.core.presentation.common.util.UiText
 import com.realityexpander.tasky.core.util.Email
 import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.emitAll
-import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.channelFlow
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.supervisorScope
 import java.time.ZonedDateTime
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
```diff
@@ -35,24 +35,38 @@ class AgendaRepositoryImpl @Inject constructor(
 
     override fun getAgendaForDayFlow(dateTime: ZonedDateTime): Flow<List<AgendaItem>> {
 
-        return flow {
+        return channelFlow {
```
philipplackner, line 38: Why channelFlow?

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
```diff
@@ -35,24 +35,38 @@ class AgendaRepositoryImpl @Inject constructor(
...
+                        // Emit the current state from DB
+                        eventRepository.getEventsForDayFlow(dateTime).collect { events ->
+                            send(events)
                         }
-                    } catch (e: CancellationException) {
-                        throw e
-                    } catch (e: Exception) {
+                    }
+                    catch (e: CancellationException) {
+                        /* intentionally eat this exception */
```
philipplackner, line 63: Why?

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
realityexpander: ## FOLLOW UP QUESTION 1

Because when i tried to `emit` values, I was getting a run-time error with `flow` about incompatible types. 

The error suggested using `channelFlow` instead. I replaced the `emit` with `send` and now it works as expected.

Not sure what else could be wrong here, do you have suggestions about how to make the normal `flow` work correctly in this situation?

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
realityexpander: Im still a little confused when to catch `CancellationExceptions`. 

In this particular case figured that ignoring a `CancellationException` would be appropriate, as the calling source doesnt need to know it was cancelled and will be handled by the `supervisorScope`

## FOLLOW UP QUESTION 2

Should the `CancellationException` be allowed to propagate to the `supervisorScope`?

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
philipplackner: Usually a channel flow is used if the flow emits faster than the collector can consume (since it comes with a buffer). But, I don't see yet why this would be necessary here. One question I have is why you launched a new coroutine (using `launch`) instead of directly executing the suspend functions in the flow?

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
philipplackner: You never really want to catch cancellation exceptions (unless you need the info that the coroutine was cancelled). In all other cases, the framework uses them to properly handle cancellation

realityexpander: OK, which video link was that? Ive seen most if not all of your coroutine videos, but I may need a refresher.

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
realityexpander: ## FOLLOW UP QUESTION 4

I thought it was needed for the Dispatcher Context for database operations. 

Maybe Room already handles that?

philipplackner: 1. Yes, it should always be the responsibility of the suspend function to switch to the dispatcher it needs (called making it main-safe - safe to call from the main dispatcher). In this case it's the responsibility of the Room/Retrofit call to switch.
2. If you don't trust them (or a 3rd party lib let's say), you should better switch using `withContext(Dispatchers.IO)` which won't launch a new coroutine, but rather just switch the dispatcher of the current one.

realityexpander: OK that makes more sense. I didnt realize you could switch coroutine contexts like that. Thanks!

PR #33, opened 2022-11-28
0031 time needle

realityexpander:
* Implement time needle.

## Question 1
Do we need to show the needle on the days that are not today? Like all the before and after days? Or does the time needle only appear on the "Now" 24-hour period?

### DIFF
--- a/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaScreen.kt
+++ b/app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaScreen.kt
@@ -442,8 +442,19 @@ fun AgendaScreenContent(
                 .fillMaxSize()
                 .padding(start = DP.tiny, end = DP.tiny)
         ) {
-            itemsIndexed(items = agendaItems) { index, agendaItem ->
+            val sortedAgendaItems = agendaItems.sortedBy { agendaItem ->
+                agendaItem.startTime
+            }
+            val agendaItemsBeforeNow = sortedAgendaItems.filter { agendaItem ->
+                agendaItem.startTime.isBefore(ZonedDateTime.now())
+            }
+            val agendaItemsAfterNow = sortedAgendaItems.filter { agendaItem ->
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaScreen.kt
```diff
@@ -484,7 +495,81 @@ fun AgendaScreenContent(
                     }
                 }
 
-                if (index < agendaItems.size - 1) {
+                if (index < agendaItemsBeforeNow.size - 1) {
+                    Spacer(modifier = Modifier.smallHeight())
+                }
+            }
```
philipplackner, line 501: Alternatively, you can use `verticalArrangement = Arrangement.spacedBy(X.dp)` which will already consider this (only works if you want to apply this spacing to the whole LazyColumn

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaScreen.kt
```diff
@@ -484,7 +495,81 @@ fun AgendaScreenContent(
...
+
+            // • TIME NEEDLE
+            item(true) {
+                Box(
+                    modifier = Modifier
+                        .fillMaxWidth()
+                        .padding(0.dp)
```
philipplackner, line 508: Not needed 🤔

philipplackner (commented):
Not sure if I get your question right 🤔  
The time needle should just show for the current day

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaScreen.kt
realityexpander: Since the Time Needle is not using spacing, I will keep the current implementation. Nice to know that the `spaceBy` works for lazylist, like a `spaceBetween`

PR #36, opened 2022-11-30
0033 refactor repo use sep get update

realityexpander:
* Separate out the `get` and `update` into two repo functions

## QUESTION 1
Im still having issues with using `emit` and `emitAll` with my setup:

https://github.com/realityexpander/Tasky/blob/43730d876d24f31883a8899867cb04c3992126ef/app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt#L82-L87

Is there something fundamental that I've set up wrong to make `send` work properly and `emit` and `emitAll` do not work?

### DIFF
--- a/app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
+++ b/app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
@@ -75,7 +75,6 @@ class AgendaRepositoryImpl @Inject constructor(
 //        return flow { // why doesnt this work?
         return channelFlow {
             supervisorScope {
-                launch(Dispatchers.IO) {
                     try {
                         updateAgendaForDayFromRemote(dateTime)
 
@@ -101,8 +100,6 @@ class AgendaRepositoryImpl @Inject constructor(
                     }
 
                 }
-
-            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
```diff
@@ -38,59 +38,68 @@ class AgendaRepositoryImpl @Inject constructor(
         return events + tasks // + reminders
     }
 
+    override suspend fun updateAgendaForDayFromRemote(dateTime: ZonedDateTime) {
+        CoroutineScope(Dispatchers.IO).launch {
```
philipplackner, line 42: I think I commented on this before, but this scope shouldn't be here. You're inside a suspend function, so you can safely execute your I/O functions. Otherwise, you also won't have any way for a caller to know when this updating finished, since nothing will suspend here.

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepostoryImpl.kt
```diff
@@ -38,59 +38,68 @@ class AgendaRepositoryImpl @Inject constructor(
...
+                    val result2 =
+                        taskRepository.upsertTaskLocally(task.toDomain())
+                    if(result2 is ResultUiText.Error) {
+                        throw IllegalStateException(result2.message.asStrOrNull())
+                    }
+                }
+            } catch (e: Exception) {
+                e.printStackTrace()
+                // don't send error to user, just log it (silent fail is ok here)
+            }
```
philipplackner, line 69: Same cancellation issue as we talked about

PR #38, opened 2022-12-03
0035 Implement offline sync

realityexpander:
This was a big PR and def all over the place because sync touches a lot of areas!

* Convert from on-line only to offline with sync
* Small UI Changes for `Time Needle`
* Added SyncRepository
* Modified many repos, api's
* Added `syncRepo` repository to track offline changes
* Removed `isDeleted` from Entities
* Added `isSynced` to Entities (prevents deletion of un-uploaded items upon refresh)
* Implemented `syncAgenda` api
* Modified error handling to fail silently when sync cache can be updated
* Sync offline changes to server
* Refactor for functionality

## Still to implement:
* Pull-to-refresh
* background sync with WorkManager

## QUESTION 1
Maybe this is better for the call, but wondering how much its worth it to make base viewmodels and repos, and trade offs for adding a heirarchy instead of a bit of repeated code. Loose Coupling vs Cohesion/Heirarchy?


app/src/main/java/com/realityexpander/tasky/MainActivity.kt
```diff
@@ -46,80 +46,58 @@ class MainActivity : ComponentActivity() {
...
-                                } else {
-                                    LoginScreenDestination
-                                },
-                            )
+                        // Check for errors
+                        if (splashState.error != null) {
+                            Toast.makeText(context, splashState.error, Toast.LENGTH_LONG).show()
+                            Thread.sleep(1000)
+                            viewModel.onSetAuthInfo(null)
                         }
```
philipplackner, line 88: This should go in a `LaunchedEffect` as well, since these functions aren't composables

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/common/convertersDTOEntityDomain/ReminderConverters.kt
```diff
@@ -17,7 +17,7 @@ fun AgendaItem.Reminder.toEntity(): ReminderEntity {
         description = description,
         remindAt = remindAt,
         time = time,
-        isDeleted = isDeleted,
+        isSynced = isSynced
```
philipplackner, line 20: Don't you also need the info how it should be synced if it's not? Meaning if it needs to be created, updated or deleted?

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepositoryImpl.kt
```diff
@@ -0,0 +1,308 @@
...
+    private val authRepository: IAuthRepository
+) : IAgendaRepository {
+
+    ///////////////////////////////////////////////
+    // • AGENDA
+
+    override suspend fun getAgendaForDay(dateTime: ZonedDateTime): List<AgendaItem> {
+        val events = eventRepository.getEventsForDay(dateTime)
+        val tasks = taskRepository.getTasksForDay(dateTime)
+        val reminders = reminderRepository.getRemindersForDay(dateTime)
```
philipplackner, line 43: These calls are all independent of each other, so they can be executed in parallel:
```
return supervisorScope {
    val events = async { eventRepository.getEventsForDay(dateTime) }
    // ...
    events.await() + tasks.await() + reminders.await()
}
```

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepositoryImpl.kt
```diff
@@ -0,0 +1,308 @@
...
+            } else {
+                Result.failure(Exception("Error getting agenda"))
+            }
+        } catch (e: Exception) {
+            Result.failure(Exception("Error getting agenda"))
+        }
+    }
+
+    override suspend fun updateLocalAgendaForDayFromRemote(dateTime: ZonedDateTime) {
+        CoroutineScope(Dispatchers.IO).launch {
```
philipplackner, line 69: The scope should be removed here, in most cases you just want to expose suspend functions and execute them in `viewModelScope`, since otherwise you have no control over the execution of this function (for example if you want to cancel it or know when it finished)

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepositoryImpl.kt
```diff
@@ -0,0 +1,308 @@
...
+                e.printStackTrace()
+                // don't send error to user, just log it (silent fail is ok here)
+            }
+        }
+    }
+
+    override fun getAgendaForDayFlow(dateTime: ZonedDateTime): Flow<List<AgendaItem>> {
+
+        return flow {
+            supervisorScope {
```
philipplackner, line 118: This is not needed here, since you're not launching child coroutines

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepositoryImpl.kt
```diff
@@ -0,0 +1,308 @@
...
+                } catch (e: Exception) {
+                        e.printStackTrace()
+                        // don't send error to user, just log it (silent fail is ok here)
+                    }
+                }
+        }
+    }
+
+    override fun addAgendaItems(agendaItems: List<AgendaItem>) {
+        CoroutineScope(Dispatchers.IO).launch {
```
philipplackner, line 138: Same comment as before

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepositoryImpl.kt
```diff
@@ -0,0 +1,308 @@
...
+        }
+
+        // Sync the `Create` API calls first
+        syncItems.forEach { syncItem ->
+            when (syncItem.modificationTypeForSync) {
+                ModificationTypeForSync.Created -> {
+                    when (syncItem.agendaItemTypeForSync) {
+                        AgendaItemTypeForSync.Event -> {
+                            val event = getEvent(syncItem.agendaItemId, true) ?: return@forEach
+                            createEvent(event, true)
```
philipplackner, line 183: I'd always use named parameters when possible, since for PR reviewers there's no way to know what this `true` stands for

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/eventRepository/eventRepositoryImpls/EventRepositoryImpl.kt
```diff
@@ -7,28 +7,43 @@ import com.realityexpander.tasky.agenda_feature.data.common.convertersDTOEntityD
...
     // • CREATE
 
-    override suspend fun createEvent(event: AgendaItem.Event): ResultUiText<AgendaItem.Event> {
+    override suspend fun createEvent(event: AgendaItem.Event, isRemoteOnly: Boolean): ResultUiText<AgendaItem.Event> {
         return try {
-            eventDao.createEvent(event.toEntity())  // save to local DB first
+            if(!isRemoteOnly) {
+                eventDao.createEvent(event.copy(isSynced = false).toEntity())  // save to local DB first
+                syncRepository.addCreatedSyncItem(event)
+            }
```
philipplackner, line 35: In which case would you not want to save it locally?

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/syncRepository/syncRepositoryImpls/SyncRepositoryImpl.kt
```diff
@@ -0,0 +1,175 @@
...
+            deletedEventIds =
+            this
+                .filter {
+                    it.agendaItemTypeForSync == AgendaItemTypeForSync.Event
+                            && it.modificationTypeForSync == ModificationTypeForSync.Deleted
+                }
+                .map {
+                    it.agendaItemId
+                },
+            deletedTaskIds =
```
philipplackner, line 155: This formatting makes it super hard to read IMO, would make sense here to store each list in a variable before for readability 🤔

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
```diff
@@ -106,6 +103,27 @@ class AgendaViewModel @Inject constructor(
...
+            // Fetch/Refresh the previous and coming week's agenda items
+            withContext(Dispatchers.IO) {
+                (-5..5).map { index ->
+                    if (index != 0) {
+                        val date = getDateForSelectedDayIndex(selectedDate, index)
+                        async { agendaRepository.getAgendaForDayFromRemote(date) }
+                    } else {
+                        async { null }
+                    }
+                }.awaitAll()
```
philipplackner, line 115: One thing to note about `awaitAll()` is that it fails as soon as one of the coroutines fails. To prevent that, you'd need another map block:
```
.map { it.await() }
```

philipplackner (approved):

philipplackner:
We can talk about the BaseViewModel thing in the call, I'm personally not a friend of it though :)

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/common/convertersDTOEntityDomain/ReminderConverters.kt
realityexpander: ## FOLLOW UP QUESTION 1

I am tracking the "modificationType" in the Sync repo. I only need to keep track of the change being sent to the server.

It seems to be working well, but maybe there is something I'm missing?

app/src/main/java/com/realityexpander/tasky/MainActivity.kt
realityexpander: ## FOLLOW UP QUESTION 2

Is the reason for putting these in a Launch block for performance? Or is there another reason to put these non-composable functions into a `Launchedeffect` block?

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepositoryImpl.kt
realityexpander: Those are calls to DAO's and only return the local items.

## FOLLOW UP QUESTION 3

I thought that because the database calls are already coroutines and accessing the same database, that there is no need to parallelize their calls.

Do database calls need to be parallelized?

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
realityexpander: Interesting. OK, I will refactor and make notes

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/syncRepository/syncRepositoryImpls/SyncRepositoryImpl.kt
realityexpander: OK, I will improve the formatting

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepositoryImpl.kt
realityexpander: ok i will refactor

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepositoryImpl.kt
realityexpander: ok i will refactor

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepositoryImpl.kt
realityexpander: will refactor this out

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/eventRepository/eventRepositoryImpls/EventRepositoryImpl.kt
realityexpander: When the app is syncing with the server, i only want to `create` an item on the server, and not locally (the app already has the item locally).

When the app is online, it will attempt to save to the server during `create`.

Is there something I'm missing here?

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/common/convertersDTOEntityDomain/ReminderConverters.kt
philipplackner: No that's fine

app/src/main/java/com/realityexpander/tasky/MainActivity.kt
philipplackner: It's just that composables should be side-effect free. In this particular case it probably won't cause any harm, but in general composable functions can execute uncontrolled in any order which is why you want to avoid placing non-compose code in them. LaunchedEffect takes care of exactly that by providing a scope that is cancelled and relaunched when the given state changes

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepositoryImpl.kt
philipplackner: Yes because in your example the call to `taskRepository.getTasksForDay(dateTime)` will only be executed when the previous event retrieval finished. All that is executed in the same suspend function, so it gets executed sequentially. Internally, DBs will surely do a lot of multi-threading but only for one given call. Here your DAO can't know that there'll be a future call to getTasks

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/eventRepository/eventRepositoryImpls/EventRepositoryImpl.kt
philipplackner: Ah okay, yeah you can do it like that. Feels a bit of a hassle to me to save one DB call while adding extra params and so, but it's fine

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/agendaRepository/agendaRepositoryImpls/AgendaRepositoryImpl.kt
realityexpander: OK, didnt know you could do this!

I will refactor

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/eventRepository/eventRepositoryImpls/EventRepositoryImpl.kt
realityexpander: Yeah I considered making separate calls, but it was a toss up.

PR #43, opened 2022-12-10
0040-add-alarms-notifications

realityexpander:
* Add alarms
* Add `remindAt` notifications
* Alarms work if the app is in background of quit
* Add card to notifications
* Add check for internet before attempting calls
* Add broadcast receivers for Alarms 
* Add abiity to complete Task from Notification
* improve readme
* add translated strings

## QUESTION 1 - App UX design
Are only `remindAt` notifications sent? 
Do we need to make `startTime` notifications too?


app/src/main/java/com/realityexpander/tasky/core/util/ConnectivityObserver/ConnectivityObserverImpl.kt
```diff
@@ -17,6 +20,61 @@ class ConnectivityObserverImpl(
 
     // use connectivityManager.requestNetwork(networkRequest, networkCallback) for api levels lower than 24
 
+    companion object {
```
MKiperszmid, line 23: Is everything in this class really useful tho?

Like, do we care if its GSM, Metered, Vpn, ethernet, etc?

Philipp has a video where he shows how to observe the network: https://youtu.be/TzV0oCRDNfM

app/src/main/java/com/realityexpander/tasky/core/presentation/notifications/RemindAtAlarmNotificationManager.kt
```diff
@@ -0,0 +1,287 @@
...
+import com.realityexpander.tasky.core.util.toUtcMillis
+import com.realityexpander.tasky.core.util.toZonedDateTime
+import logcat.logcat
+import java.time.ZonedDateTime
+import java.time.format.DateTimeFormatter
+import java.util.*
+
+
+
+interface IRemindAtAlarmNotificationManager {
```
MKiperszmid, line 38: Usually the interfaces are declared on the domain package, and have their implementations on the data layer, since usually the interface is used by something on the domain layer

app/src/main/java/com/realityexpander/tasky/core/presentation/notifications/RemindAtAlarmManager.kt
```diff
@@ -0,0 +1,185 @@
...
+                    futureItems.map { agendaItem ->
+                        when (agendaItem) {
+                            is AgendaItem.Event -> agendaItem.title
+                            is AgendaItem.Task -> agendaItem.title
+                            is AgendaItem.Reminder -> agendaItem.title
+                            else -> {
+                                "UNKNOWN_AGENDA_ITEM_TYPE"
+                            }
+                        }
+                    }.toList().toCollection(ArrayList())
```
MKiperszmid, line 81: Do you really need to do a .toList() and then .toCollection(ArrayList())?

app/src/main/java/com/realityexpander/tasky/core/presentation/notifications/RemindAtAlarmManager.kt
```diff
@@ -0,0 +1,185 @@
...
+            REMIND_AT_ALARM_SUPERVISOR_REQUEST_CODE,
+            Intent(context, MainActivity::class.java).also {
+                it.putExtra(
+                    CURRENT_ALARM_PENDING_INTENTS,
+                    arrayOf<PendingIntent>(*alarmPendingIntents.toTypedArray())
+                )
+                it.putStringArrayListExtra( // for debugging
+                    CURRENT_ALARM_TITLES,
+                    futureItems.map { agendaItem ->
+                        when (agendaItem) {
```
MKiperszmid, line 73: Cant you just do agendaItem.title ?

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
```diff
@@ -132,6 +132,25 @@ class AgendaViewModel @Inject constructor(
...
+        // Set Alarms for Agenda Items for coming week
+        viewModelScope.launch {
+
+            agendaRepository.getLocalAgendaItemsRemindAtForDateTimeRangeFlow(
+                ZonedDateTime.now().truncatedTo(ChronoUnit.DAYS),
+                ZonedDateTime.now().truncatedTo(ChronoUnit.DAYS).plusWeeks(1)
+            )
+                .collect { agendaItems ->
+                    //viewModelScope.launch {
+                    delay(500)
```
MKiperszmid, line 145: Why do we have a delay?

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
```diff
@@ -132,6 +132,25 @@ class AgendaViewModel @Inject constructor(
...
+                        _oneTimeEvent.emit(OneTimeEvent.SetAllAgendaItemAlarms(agendaItems))
+                    //}
+                }
+        }
+
+//        viewModelScope.launch {
+//            _oneTimeEvent.tryEmit(OneTimeEvent.SetAllAgendaItemAlarms(agendaItems))
```
MKiperszmid, line 152: This comment can be removed

app/src/main/java/com/realityexpander/tasky/agenda_feature/domain/AgendaItem.kt
```diff
@@ -40,11 +42,13 @@ abstract class AgendaItem {
         override val id: UuidStr,
         override val title: String,
         override val description: String,
-        val remindAt: ZonedDateTime,
         val isSynced: Boolean = false,
 
         val time: ZonedDateTime,
         override var startTime: ZonedDateTime = time, // for sorting in Agenda
+        val remindAt: ZonedDateTime,
+        override var remindAtTime: ZonedDateTime = remindAt,
```
MKiperszmid, line 50: Whats the different between remindAt and remindAtTime?

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/eventRepository/eventRepositoryImpls/EventRepositoryImpl.kt
```diff
@@ -1,4 +1,4 @@
-package com.realityexpander.tasky.agenda_feature.data.repositories.eventRepository.eventRepositoryImpls
+package com.realityexpander.eventy.agenda_feature.data.repositories.eventRepository.eventRepositoryImpls
```
MKiperszmid, line 2: Did you rename the app to Eventy?

Or did you rename Task to Event and it got renamed by mistake?

MKiperszmid (commented):
Left some comments.

Also, try to stick to smaller PRs. Try reaching a max of 400-500 lines. Else, split the PR

app/src/main/java/com/realityexpander/tasky/core/util/ConnectivityObserver/ConnectivityObserverImpl.kt
realityexpander: Im using his observer here. 

I was thinking I would have to support all those others but I just made a companion object with a simple bash script to ping the network for internet availability in some parts to avoid making unnecessary calls.

app/src/main/java/com/realityexpander/tasky/core/presentation/notifications/RemindAtAlarmManager.kt
realityexpander: No, i can just to `.toCollection()`

Is there a better way to create an ArrayList from a List?

app/src/main/java/com/realityexpander/tasky/core/presentation/notifications/RemindAtAlarmManager.kt
realityexpander: YES!

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
realityexpander: I changed it to a debounce

app/src/main/java/com/realityexpander/tasky/agenda_feature/domain/AgendaItem.kt
realityexpander: remindAt is represented as a `Long` in the DTO and Entity classes. So to make it clearer and easier to work with, i added a second property to leave the more human-readable form of the ZonedDateTime (for debugging)

## FOLLOW UP QUESTION 2

Let me know if there is a better approach.

app/src/main/java/com/realityexpander/tasky/agenda_feature/data/repositories/eventRepository/eventRepositoryImpls/EventRepositoryImpl.kt
realityexpander: Whoops, I will fix it.

realityexpander:
> Left some comments.
> 
> Also, try to stick to smaller PRs. Try reaching a max of 400-500 lines. Else, split the PR

Yes, its something I am currently struggling with

app/src/main/java/com/realityexpander/tasky/core/presentation/notifications/RemindAtAlarmNotificationManager.kt
realityexpander: OK, I have moved them to the `domain` for `agenda_feature`

app/src/main/java/com/realityexpander/tasky/core/presentation/notifications/RemindAtAlarmManager.kt
philipplackner: Looks good to me

app/src/main/java/com/realityexpander/tasky/agenda_feature/domain/AgendaItem.kt
philipplackner: Yeah I agree that having 2 so similar properties with the same type is quite confusing 🤔 
I think one property is fully enough here and if you see `remindAt` is of type `ZonedDateTime` this makes it fully clear to me

app/src/main/java/com/realityexpander/tasky/core/presentation/notifications/RemindAtAlarmManager.kt
philipplackner: Looks good to me

philipplackner:
Yes, notifications are only for remindAt

app/src/main/java/com/realityexpander/tasky/agenda_feature/domain/IRemindAtAlarmManager.kt
```diff
@@ -0,0 +1,16 @@
...
+
+    fun setAlarmsForAgendaItems(
+        context: Context,
+        agendaItems: List<AgendaItem>
+    )
+
+    fun cancelAllAlarms(
+        context: Context,
+        onFinished: () -> Unit = {},
+    )
```
philipplackner, line 15: Instead of providing the context as parameter in the domain layer (which is Android-specific and therefore not isolated), I'd pass it to the constructor of the class implementing `IRemindAtAlarmManager`.
Otherwise, you can't create fakes of this interface for local JVM tests.

app/src/main/java/com/realityexpander/tasky/agenda_feature/domain/IRemindAtNotificationManager.kt
```diff
@@ -0,0 +1,14 @@
...
+import android.content.Intent
+
+interface IRemindAtNotificationManager {
+
+    fun showNotification(
+        context: Context,
+        alarmIntent: Intent
+    )
+
+    fun createNotificationChannel(context: Context)
```
philipplackner, line 13: Same here

app/src/main/java/com/realityexpander/tasky/core/presentation/notifications/RemindAtAlarmManager.kt
```diff
@@ -0,0 +1,166 @@
...
+import com.realityexpander.tasky.core.presentation.notifications.RemindAtNotificationManager.ALARM_NOTIFICATION_INTENT_ACTION_ALARM_TRIGGER
+import com.realityexpander.tasky.core.presentation.notifications.RemindAtNotificationManager.ALARM_NOTIFICATION_INTENT_EXTRA_AGENDA_ITEM
+import com.realityexpander.tasky.core.presentation.notifications.RemindAtNotificationManager.ALARM_NOTIFICATION_INTENT_EXTRA_ALARM_ID
+import com.realityexpander.tasky.core.util.toUtcMillis
+import com.realityexpander.tasky.core.util.toZonedDateTime
+import logcat.logcat
+import java.time.ZonedDateTime
+import java.time.temporal.ChronoUnit
+
+object RemindAtAlarmManager : IRemindAtAlarmManager {
```
philipplackner, line 21: This should be a class taking the app context as a parameter. You can then provide it with Hilt

app/src/main/java/com/realityexpander/tasky/core/presentation/notifications/RemindAtNotificationManager.kt
```diff
@@ -0,0 +1,332 @@
...
+import com.realityexpander.tasky.core.util.UuidStr
+import com.realityexpander.tasky.core.util.toIntegerHashCodeOfUUIDString
+import com.realityexpander.tasky.core.util.toUtcMillis
+import com.realityexpander.tasky.core.util.toZonedDateTime
+import logcat.logcat
+import java.time.format.DateTimeFormatter
+import java.util.*
+
+
+object RemindAtNotificationManager : IRemindAtNotificationManager {
```
philipplackner, line 41: Same here

app/src/main/java/com/realityexpander/tasky/core/presentation/notifications/RemindAtNotificationManager.kt
```diff
@@ -0,0 +1,332 @@
...
+            "🔘 $title\n" +
+                    "• Starting at ${
+                        startDateTimeUtcMillis.toZonedDateTime()
+                            .format(DateTimeFormatter.ofPattern("h:mm a, E MMM d"))
+                    }" + "\n\n" +
+                    "• $description",
+            20f, 70f + offsetY2ForTask, width, height, paint
+        )
+        return bitmap
+    }
```
philipplackner, line 305: Just curious, what does this draw? 😁

philipplackner (approved):

app/src/main/java/com/realityexpander/tasky/agenda_feature/domain/AgendaItem.kt
realityexpander: I guess Im not understanding what to do here.

I have a `remindAt` which is a `ZonedDateTime` in the domain layer, but its a `Long` represending `EpochSeconds` in the data layer. Also, when debugging the data layer, the `ZonedDateTime` are easier to work with (shows up as a human readable in the debugger) and are marked `@Transient`, so the data layer never stores it or sends it.

I hope this makes more sense why I am using two variables here.

app/src/main/java/com/realityexpander/tasky/core/presentation/notifications/RemindAtNotificationManager.kt
realityexpander: This draws the card for the notification, like this:

[](https://user-images.githubusercontent.com/5157474/206991659-fcd25421-6011-43db-a853-bfa4cbb3311d.png)

app/src/main/java/com/realityexpander/tasky/agenda_feature/domain/IRemindAtAlarmManager.kt
realityexpander: OK, i will refactor to use hilt

app/src/main/java/com/realityexpander/tasky/core/presentation/notifications/RemindAtNotificationManager.kt
realityexpander: will do

app/src/main/java/com/realityexpander/tasky/agenda_feature/domain/IRemindAtNotificationManager.kt
realityexpander: will do

app/src/main/java/com/realityexpander/tasky/core/presentation/notifications/RemindAtNotificationManager.kt
MKiperszmid: Ahh so you created a custom design for the Notification

Looks good, but im curious tho

What would happen if the title is too big or too small? Same with description
Ex; Description has 2k characters

app/src/main/java/com/realityexpander/tasky/agenda_feature/domain/AgendaItem.kt
MKiperszmid: But in this case, both are of type ZonedDateTime

And it says: remindAtTime = remindAt

Thats what im not fully understanding

philipplackner: Yes that's the same I don't understand. Nothing wrong with storing it as ZonedDateTime, but why have 2 same properties?

realityexpander: OK, i could store them as ZonedDateTime strings.

The payload coming from the API is in `Epoch` `Long`

I want to be able to check if the date is correct and human readable, which is why I copy the `remindAt` time to a ZonedDateTime and keep it for debugging purposes. I was finding myself constantly using the web sites to convert `EpochMillis` to `ZonedDateTime` format, and this was an easy way to reduce that.

I hope this makes sense.

philipplackner: It would be new to me that it could do that without type converters, but if you say it can compare the other way around why would it not be able to do it 🤔 
Either way, regarding the issue you usually have an entities model in data which contains the fields stored in your DB and you map that to your domain model when reading it. Why does the domain model then need 2 date time fields?

realityexpander: I need two because one of the fields (the `EpochSeconds Long` and `EpochMillis Long`) are not human readable, and I needed to use a conversion app to see what the dates and times actually are. So instead of having to use the conversion website to see what the dates are for these `Longs`, I just used another var of type `ZoneDateTime` so its very easy to see what the date is without resorting to using an app.

How do you convert `long` `EpochMillis` and `EpochSeconds` to human readable when you are debugging?

philipplackner: If I just want to check which date is arriving from the API, I inspect the dateTime value of the domain model to see that.  I think what would make more sense in this case is to give `AgendaItem` a constructor and put the abstract values in there (without abstract). That way, the child classes only have one field per type.
It's quite confusing if you see `remindAt` and `remindAtTime` as properties for an event, since they're effectively the same.

realityexpander: How do you overload the data stored in the Room Database? AFAIK you can only store a single type in the Entity. Pretty confused about what is being suggested here.

For dates and times, The domain layer uses `ZonedDateTime`. The data layer (room) MUST use `long` to represent the date/time. This is what is stored in the database and what Room uses for comparisons using SQL queries.

Please clarify about what is to be stored where and what types:

ie: data layer -> `long` representing Epoch Seconds for database, `long` representing Epoch Millis for the API
     domain layer -> `ZonedDateTime`
     presentation layer -> `ZonedDateTime`

Im missing something important here, and cant figure out what it is yet.

MKiperszmid: "Using the above setup, it's not possible to visually just look at the data and know the date using the Database inspector by looking at the raw data at the column. Same for the API calls looking at the raw data in and out. You must rely on the domain layer to see the data is correct. If this is the way it works, then I will leave it be."

Since the API and DB use epoch seconds (Long), then no, a human won't fully understand the date just by looking at it, and you'd need to use a site like https://www.epochconverter.com/ to understand it. That's why as soon as you receive it from either DB or API, you convert it to a ZonedDateTime on your Domain model. So it's easy to read and understand

You should only use Long (seconds) when storing to API and DB, and the reason why is because if you store as Date, each person understands date a different way. That's why there are many ways to say a specific date (month/day/year, day/month/year, day-month-year, etc). But if you just store it as seconds, then when you parse it to ZonedDateTime, you know the date is 100% correct

realityexpander: OK, this is good confirmation. Thank you!

PR #45, opened 2022-12-13
0042 refactor DI app module to separate modules

realityexpander:
* Clean up AppModule


app/src/main/java/com/realityexpander/tasky/di/NetworkingModule.kt
```diff
@@ -0,0 +1,159 @@
...
+
+                // Check for a valid AuthToken in the IAuthApi Companion object.
+                //   If not valid, attempt to get it from the AuthDao.
+                //   If valid, set it in the IAuthApi Companion object, for faster access.
+                IAuthApi.authToken ?: run {
+                    val authToken = authDao.getAuthToken() // could take a while.
+                    if(authToken != null) {
+                        IAuthApi.authToken = authToken
+                    }
+                }
```
philipplackner, line 81: Could it make sense to extract this in a function part of `IAuthApi`? So that callers just need to call `IAuthApi.getToken()` which will fetch it if it doesn't exist or instantly return it if it does?

realityexpander: OK I will refactor for this.

PR #46, opened 2022-12-13
0043 - Refactor to start workers in the `AgendaViewModel`

realityexpander:


app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaScreen.kt
```diff
@@ -318,6 +323,17 @@ fun AgendaScreenContent(
...
+            OneTimeEvent.StartWorkers -> {
+                logcat { "OneTimeEvent.StartWorkers: starting Tasky workers" }
+                SyncAgendaWorker.startWorker(context)
+                RefreshAgendaWeekWorker.startWorker(context)
+            }
+            OneTimeEvent.StopWorkers -> {
+                logcat { "OneTimeEvent.StopWorkers: stopping Tasky workers" }
+                WorkManager.getInstance(context).cancelAllWorkByTag(TASKY_WORKERS_TAG)
+            }
+
```
philipplackner, line 336: Where do you send these events? Why does this need to be triggered from the UI and not when the VM is initialized?

philipplackner (approved):

app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaScreen.kt
realityexpander: Ha, I had it set that way, and @MKiperszmid asked me to switch it to triggering it in the VM... 

SO now im confused on where this trigger should be kept! 

## FOLLOW UP QUESTION 1

Please coordinate with him and let me know what is the best way to handle this.

philipplackner: Just putting worker stuff in the VM will not make them locally unit testable, but you can put them behind an abstraction, that's at least what I did. My implementation for using these then sits in data.
I personally just don't see a reason in executing something in the UI that should just be done once per screen (especially because from the UI it will execute on every config change)

MKiperszmid: Hey!

Im sorry if I wasn't clear enough

What I meant was to run the workers, in the init block of the ViewModel. Not get them triggered by UI

realityexpander: The `init` block in the VM does trigger the workers to start, but the workers require a context, which is only available in the UI layer, which is why im using a one-time event sent to the UI.

Is there a way to trigger the workers to start from the VM without using context? Seems like the context is necessary to start a worker, please let me know if Im wrong here.

MKiperszmid: One way to "bypass" having to pass a context, is to get the Workmanager instance on DI using ApplicationContext
Something like:

`@Provides
    @Singleton
    fun provideWorkManager(
        application: Application
    ): WorkManager {
        return WorkManager.getInstance(application)
    }`


So you then receive WorkManager on the ViewModel's constructor, and you can use it with no context

realityexpander: ah, ok, i didnt think about doing it that way because the OS is calling the custom workmanager class, but that makes sense, i will refactor for this.

philipplackner: @MKiperszmid This still binds your ViewModel to an Android dependency (WorkManager) which does not make it locally unit testable. The idea is correct, but the fix would be putting it behind an abstraction:
```
// Domain layer
interface AgendaSyncScheduler {
    fun scheduleSync(...)
    fun cancelSync(...)
}

// Data layer
class AgendaSyncSchedulerImpl(val workManager: WorkManager): AgendaSyncScheduler {
    ...
}

// ViewModel
class YourViewModel(val syncScheduler: AgendaSyncScheduler): ViewModel() { ... }
```

MKiperszmid: Yes you are right!


As philipp says. To avoid having Android dependencies on classes you want to unit test, you should hide said Android dependency on an Abstraction

That way, you can pass a "FakeWorker" when unit testing

realityexpander: OK I will refactor for this implementation

PR #51, opened 2022-12-16
0049 refactor workers use abstraction

realityexpander:
* Workers now are scheduled thru an abstraction to allow for Unit Testing


app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaViewModel.kt
```diff
@@ -242,7 +234,7 @@ class AgendaViewModel @Inject constructor(
             }
             is Logout -> {
                 viewModelScope.launch {
-                    workManager.cancelAllWorkByTag(TASKY_WORKERS_TAG)
+                    agendaWorkersScheduler.cancelAllWorkers()
```
philipplackner, line 237: Is this also called if you're logged out due to an expired or invalid token?

realityexpander: Will refactor for this.

PR #54, opened 2022-12-16
0051 Indicate internet on login register screen

realityexpander:
* add internet connectivity indicator on login/register screen
* Refactor Keyboard Visibility observer to a function


app/src/main/java/com/realityexpander/tasky/agenda_feature/presentation/agenda_screen/AgendaScreen.kt
```diff
@@ -110,37 +111,7 @@ fun AgendaScreen(
...
-                modifier = Modifier
-                    .align(Alignment.TopCenter)
-                    .background(Color.Red.copy(alpha = .5f))
-                    .fillMaxWidth()
-                    .align(Alignment.Center),
-
-                )
-        }
-    }
+    ShowInternetAvailabilityIndicator(connectivityState)
```
philipplackner, line 114: Is this a normal composable? If yes, shouldn't it rather be called `InternetAvailabilityIndicator` since a composable function is not really an action that's done?

realityexpander: Yes this makes sense, I will rename for clarity.

PR #56, opened 2022-12-18
0053 add deep links

realityexpander:
* Allows user creating an event to pass a deep link to the event (that the user was invited to)


philipplackner (commented):
Looks good 👍

PR #58, opened 2022-12-20
0055 add transitions

realityexpander:
* Add transitions between screens


app/src/main/java/com/realityexpander/tasky/core/presentation/animatedTransitions/ScreenTransitions.kt
```diff
@@ -0,0 +1,72 @@
...
+                    slideOutHorizontally(
+                    targetOffsetX = { 1000 },
+                    animationSpec = tween(700)
+                )
+            else -> slideOutHorizontally(
+                    targetOffsetX = { -1000 },
+                    animationSpec = tween(700)
+                )
+        }
+    }
```
philipplackner, line 71: Would it make sense to save these repeating animations in a variable on top here? That way you can better change the hardcoded values and/or animations and this class becomes a bit lighter

philipplackner (approved):

app/src/main/java/com/realityexpander/tasky/core/presentation/animatedTransitions/ScreenTransitions.kt
realityexpander: OK, I will refactor

